PR #6, opened 2022-01-04
Tooltip

locothedev:
Added Tooltip component with customizable text


src/components/Tooltip/Tooltip.styles.ts
```diff
@@ -0,0 +1,49 @@
+import fonts from '../../styles/fonts';
+import resetCSS from '../../styles/reset';
+
+const initialStyles = `
+    ${resetCSS}
+    ${fonts.text}
+    background-color: #112F5C;
```
BillyG83, line 7: please add this new color to the colors.js as "blueMidnight"

src/components/Tooltip/Tooltip.styles.ts
```diff
@@ -0,0 +1,49 @@
...
+    border-radius: 5px;
+    align-items: center;
+    left: 50%;
+    bottom: calc(100% + 5px);
+    pointer-events: none;
+`
+
+const box = `
+    position: relative;
+    color: #112F5C;
```
BillyG83, line 36: use new blueMidnight here please

src/components/Tooltip/Tooltip.styles.ts
```diff
@@ -0,0 +1,49 @@
...
+    text-align: center;
+    border-radius: 5px;
+    padding: 10px 8px;
+    font-size: 14px;
```
BillyG83, line 40: if this component holds text please use fonts.text and if you need a modifier like fonts.smallPrint please add that as the text will need the font family, line-height, etc etc. TY

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,149 @@
...
+    ${tooltipStyles.container}
+    ${(p) => getContainerStyleByPosition(p.position)}
+`
+
+const TooltipBox = styled.span<Pick<TooltipProps, "position">>`
+    ${tooltipStyles.initialStyles}
+    ${tooltipStyles.box}
+
+    ${(p) => getBoxStyleByPosition(p.position)}
+    color: white;
```
BillyG83, line 27: please use color.js const white. TY

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,149 @@
...
+`
+
+const getBoxStyleByPosition = (position: "top" | "bottom" | "left" | "right" | undefined) => {
+    switch (position) {
+        case "bottom":
+            return `
+            &:after {
```
BillyG83, line 34: if its one that a line or two can you please wrap these strings in CSS consts and move them to styles.ts 
Thanks and sorry

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,149 @@
...
+        case "right":
+            return `
+                margin-left: 0;
+                width: 100%;
+                top: 50%;
+                left: calc(100% + 5px);
+                width: max-content;
+            `;
+        default:
+            return `
```
BillyG83, line 99: this one is ok as it can be one line, but the rest above maybe move them to styles please
`return bottom: calc(100% + 5px);`

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,149 @@
...
+const Tooltip: React.FC<TooltipProps> = ({
+id=String(Date.now()),
+position="bottom",
+text="Tooltip text",
+children
+}: TooltipProps) => {
+
+    const [showTooltip, setVisibility] = useState(false)
+
+    if(!children || children?.length === 0) {
```
BillyG83, line 114: this could be a little cleaner, maybe
```
{
  children || <div style={{display: 'none'}}></div>
}
```
but if there is always children needed should the prop type be options at all?

then it could just be
{ children }

src/components/Tooltip/types.ts
```diff
@@ -0,0 +1,21 @@
+export interface TooltipProps {
+    /**
+	 * The Tooltip ID will be generated if not assigned
+	 */
+	id?: string;
+
+	/**
+	 * The text in Tooltip
+	 */
+	text?: string;
```
BillyG83, line 10: I don't think this should be optional, if the dev wants a tooltip they should be prompted to add text

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,107 @@
...
+                  border-right-color: ${color.blueDark2};
+                  left: calc(${tooltipStyles.arrowSizeTooltip} * -1);
+                  top: 50%;
+                  transform: translateX(0) translateY(-50%);
+                }
+            `;
+    default:
+      return `
+                bottom: calc(100% + 5px);
+                background-color: ${color.blueDark2};
```
BillyG83, line 75: please move this and the 4 cases above it to the CSS file 🙏

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,107 @@
...
+            `;
+  }
+};
+
+const Tooltip: React.FC<TooltipProps> = ({
+  id = String(Date.now()),
+  position = 'bottom',
+  text = 'Tooltip text',
```
BillyG83, line 83: this can just be `text` if it is not optional in the Props

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,107 @@
...
+  children,
+}: TooltipProps) => {
+  const [showTooltip, setVisibility] = useState(false);
+
+  if (!children || children?.length === 0) {
```
BillyG83, line 88: if no children return nothing? Please make children a non optional property and remove this

src/components/Tooltip/Tooltip.styles.ts
```diff
@@ -0,0 +1,43 @@
...
+    justify-content: center;
+    left: 50%;
+    padding: 8px;
+    pointer-events: none;
+    position: absolute;
+    transform: translateX(-50%);
+    white-space: nowrap;
+
+    &:before {
+        border-top-color: red;
```
BillyG83, line 27: can you use colors.js please?

locothedev: @Y0moo for what are we using red here?

src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+    <Tooltip {...args} />
+  </div>
+);
+
+export const NotVisible = Template.bind({});
+NotVisible.args = {};
+
+export const Bottom = Template.bind({});
+Bottom.args = {
+  children: [<Icon key='0' svg={iconTypes.helpCircle} fill="grey" />],
```
BillyG83, line 30: please use colors.js here for fill and the other stories

src/components/Tooltip/Tooltip.styles.ts
Y0moo: removed on a new commit

locothedev:
Thanks @BillyG83 & @Y0moo

PR #7, opened 2022-01-04
Popup: Notification

locothedev:
Notification popup.
Customizable Title+Message
Customizable Icon
Closable


src/components/Notification/types.ts
```diff
@@ -0,0 +1,29 @@
...
+    
+        /**
+         * The title to display in notification
+         */
+        title?: string
+    
+        /**
+         * The message to display in notification
+         */
+        message?: string;
```
BillyG83, line 17: i think `message` should probably not be optional, i cant think of a case where we would nag a user with a notification that had no text. The Kresmir message is great but lets keep that as a Storybook easter egg so its not hard baked into our component

src/components/Notification/types.ts
```diff
@@ -0,0 +1,29 @@
...
+        /**
+         * set an icon to show inside the notification
+         * import { iconTypes } from "../../components/Icon/collection"
+         */
+        icon?: iconTypes;
+
+        /**
+         * set if notification is active/viewable
+         */
+        active?: boolean;
```
BillyG83, line 28: For this Boolean would you consider that `isVisible` or `isShowing` could be more intuitive? 

My old boss would insist that all Boolean should be isSomething or hasSomething but i get that things like disabled and visible seem more intuitive in some cases

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,68 @@
...
+            <BoxStyled>
+                <FlexStyled>
+                <ParagraphStyled data-testid={'test-notification-title'}>
+                    {title}
+                </ParagraphStyled>
+                <div
+                    data-testid={'test-notification-x'}
+                    onClick={() => setVisible(false)}
+                >
+                    <Icon size={20} svg={iconTypes.x} />
```
BillyG83, line 55: please set the `fill` value on Icons as i have a plan to change the default to `inherit` so it can inherit CSS changes like hover. This is my bad sorry

BillyG83: and use color.js please 😁

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,68 @@
...
+const Notification: React.FC<NotificationProps> = ({
+    id = String(Date.now()),
+    icon,
+    message = "Kresimir: Thank you for sharin..",
+    title = "New Message",
+    active = false,
+}: NotificationProps) => {
+    const [visible, setVisible] = useState(active);
+
+    useEffect(() => {
```
BillyG83, line 37: i could be wrong but i dont think you need this useEffect as `active` is a property supplied to this component, if `active` was to change the component would re-redner anyway

src/components/Notification/Notification.test.tsx
```diff
@@ -0,0 +1,168 @@
...
+	});
+
+	it("should render left icon", () => {
+		const iconSVG = container.querySelector(`[data-testid="${testId}"] > svg`);
+		expect(iconSVG).not.toBeNull();
+	});
+
+    it("should render close icon", () => {
+		const iconSVG = container.querySelector(`[data-testid="${closeId}"] > svg`);
+		expect(iconSVG).not.toBeNull();
```
BillyG83, line 53: this checks that there is an SVG but you can also check the textContent of the SVG and it will return its title text to make sure its the correct SVG, but that test is also in `<Icon />` so maybe its over kill 😅

src/components/Notification/Notification.styles.tsx
```diff
@@ -0,0 +1,54 @@
...
+    background: #112F5C;
+    border-radius: 20px;
+    color: white;
+    min-width: 270px;
+    max-width: 25%;
+    height: 84px;
+    padding: 16px;
+    display: flex;
+    align-items: center;
+    font-size: 14px;
```
BillyG83, line 17: I think we need ${fonts.textSmall} now right? cuz font size 14px should be save there but this will also effect the line height, check the typography on the far left of figma and make a textSmall please

src/components/Notification/Notification.styles.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+const initialStyles = `
+    ${resetCSS}
+    ${fonts.text}
+    background: #112F5C;
+    border-radius: 20px;
+    color: white;
+    min-width: 270px;
+    max-width: 25%;
+    height: 84px;
```
BillyG83, line 13: A fixed height seems dangerous, have you tested this with a long string of text? I would remove this and use padding to push the text off the edges of the component if needed. This will mean you prob dont need to flex the component either

src/components/Notification/Notification.styles.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+const message = `
+color: ${color.grey};
+font-weight: 400;
+font-family: IBM Plex Mono;
+font-size: 16px;
+font-style: normal;
+font-weight: 400;
+line-height: 24px;
+letter-spacing: 0px;
```
BillyG83, line 34: 100% needs a new fonts.js update here please 
```
${fonts.code}
```
perhaps?

src/components/Notification/Notification.styles.tsx
```diff
@@ -0,0 +1,54 @@
...
+    padding: 16px;
+    display: flex;
+    align-items: center;
+    font-size: 14px;
+`
+
+const box = `
+    display: grid;
+    width: 100%;
+    margin-left: 0.5rem;
```
BillyG83, line 23: is REM needed, i would rather not have a mix of values, lets try to stick to `px` and `%` unless you need this?

src/components/Notification/Notification.styles.tsx
```diff
@@ -0,0 +1,54 @@
...
+font-style: normal;
+font-weight: 400;
+line-height: 24px;
+letter-spacing: 0px;
+text-align: left;
+`
+
+const title = `
+    margin: 0;
+    font-weight: 700;
```
BillyG83, line 40: pretty sure we have fonts.bold, have a look or add it if you need it

src/components/Notification/types.ts
locothedev: Gotch ya!

src/components/Notification/types.ts
locothedev: isVisible sounds better

src/components/Notification/Notification.tsx
locothedev: Unfortunately I used without useEffect but switching active props in storybook didnt close the notification somehow

src/components/Notification/Notification.test.tsx
locothedev: Will change test to match iconType with svg title

src/components/Notification/Notification.styles.tsx
locothedev: Gotchya!

src/components/Notification/Notification.styles.tsx
locothedev: I 've used % . Seems I changed it for whatever reason. Will correct that

src/components/Notification/Notification.styles.tsx
locothedev: Gotchya

src/components/Notification/Notification.styles.tsx
locothedev: will change to px

src/components/Notification/Notification.styles.tsx
locothedev: Gotchya!

locothedev:
@BillyG83 done all the above mentioned.
But somehow prettier does not want to run properly on my end it seems that somewhere I'm overriding prettier settings.

PR #14, opened 2022-01-07
feat: Add crypto cards

dani69654:


src/components/CryptoCards/CryptoCards.tsx
```diff
@@ -0,0 +1,84 @@
...
+    ${cryptoCardStyles.chainNameStyle}
+`;
+
+const ImageBoxStyle = styled.div`
+    ${cryptoCardStyles.imageBoxStyle}
+`;
+
+const Logo = styled.div`
+    ${cryptoCardStyles.logo}
+`;
```
locothedev, line 43: Please move all styled consts to styles.ts and try to keep component file clean from css/styles

src/components/CryptoCards/types.ts
```diff
@@ -0,0 +1,43 @@
+import { iconTypes } from '../Icon/collection';
+
+export interface CryptoCardProps {
+    /**
+     * A function that will be called if the button is clicked
+     */
+    buttonClickEvent?: () => void;
```
locothedev, line 7: let's re-name this to prop to onClick and give it following type
`(event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void;`

src/components/CryptoCards/types.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * The type of the chain / a subtitle below the chain name
+     */
+    chainType: string;
+
+    /**
+     * The chain logo
+     */
+    chainLogo: string;
```
locothedev, line 22: I'm not sure if we should just stick to the chains we have. @BillyG83 what do you think about :

`chain: "eth" | "bsc" | "fantom" | "polygon" | "arbitrum"` etc etc.

locothedev:
@dani69654 Please add some tests before merging otherwise the pipeline fails. 

Also I think we should stick to svg instead of png files

src/components/CryptoCards/CryptoCards.tsx
```diff
@@ -0,0 +1,84 @@
...
+    chain,
+    chainType,
+    chainLogo,
+    settingsIcon,
+    settingsColor,
+    bgColor,
+    btnText,
+}: CryptoCardProps) => {
+    return (
+        <CryptoCardStyled style={{ background: bgColor || 'black' }}>
```
locothedev, line 56: Try to avoid inline css

src/components/CryptoCards/CryptoCards.styles.tsx
```diff
@@ -0,0 +1,177 @@
+import resetCSS from '../../styles/reset';
+import fonts from '../../styles/fonts';
+import color from '../../styles/colors';
+import { css } from 'styled-components';
+import { getShade } from '../../styles/colors';
+
+const borderStyle = css`
+    ${resetCSS}
+    height: 336px;
+    width: 265px;
```
locothedev, line 10: Do not use fixed sizes for height/width

src/components/CryptoCards/CryptoCards.styles.tsx
```diff
@@ -0,0 +1,177 @@
...
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+    align-items: center;
+    padding: 24px 40px;
+    position: absolute;
+    width: 265px;
+    height: 336px;
+    left: 20px;
+    top: 20px;
```
locothedev, line 27: Redudant css code

src/components/CryptoCards/CryptoCards.styles.tsx
BillyG83: prop you want width 100%, max with 256, display block.... in this case you maybe do want a set height as they must all stand in a row and we dont want one bigger than the rest... however we could use flex on the parent element for these. fixed height is ok for now

src/components/CryptoCards/CryptoCards.styles.tsx
BillyG83: I'm not sure why this is redundant, let me check the MR fully, but i'm sure Loco is correct. One thing i would say is we are using a VSCode plugin called 'sort lines' we use this to alphabetise props in CSS and JS, if you could do that to align, i would be vert grateful. TY

locothedev: Look at padding for example its there twice

src/components/CryptoCards/types.ts
dani69654: done

src/components/CryptoCards/types.ts
dani69654: done

src/components/CryptoCards/CryptoCards.tsx
dani69654: done

src/components/CryptoCards/CryptoCards.styles.tsx
dani69654: done

src/components/CryptoCards/CryptoCards.styles.tsx
dani69654: done

src/components/CryptoCards/CryptoCards.tsx
```diff
@@ -29,27 +29,27 @@ const CryptoCards: React.FC<CryptoCardProps> = ({
...
-                    </ChainNameStyle>
-                    <NetworkStyle>{chainType || 'Network'}</NetworkStyle>
-                </NetworkInfoHolder>
-            </InfoStyle>
-            <ButtonStyles onClick={onClick} data-testid={'test-button'}>
+                    </PChainNameStyled>
+                    <PNetworkStyled>{chainType || 'Network'}</PNetworkStyled>
+                </DivNetworkInfoStyled>
+            </DivInfoStyled>
+            <ButtonStyled onClick={onClick} data-testid={'test-button'}>
```
locothedev, line 45: See on slack

src/styles/colors.ts
```diff
@@ -1,4 +1,5 @@
 const color = {
+    black: '#000000',
```
dani69654, line 2: I should have done this in a different branch.

src/styles/fonts.ts
```diff
@@ -33,6 +33,11 @@ const h4 = `
   line-height: 28px;
 `;
 
+const h5 = `
+    font-size: 16px;
+    line-height: 25px;
+`;
+
```
dani69654, line 40: I should have done this in a different branch.

src/styles/fonts.ts
```diff
@@ -51,6 +56,10 @@ const text = `
   line-height: 24px;
 `;
 
+const textBold500 = `
+  font-weight: 500;
+`;
+
```
dani69654, line 62: I should have done this in a different branch.

src/styles/fonts.ts
```diff
@@ -71,18 +80,30 @@ const semiBold = `
...
 `;
 
+const openSans = `
+    font-family: Open Sans;
+`;
+
+const montserrat = `
+    font-family: Montserrat;
+`;
+
```
dani69654, line 90: I should have done this in a different branch.

PR #18, opened 2022-01-07
feat: get percentage by color

locothedev:
You can get the percentage of a color now like this

First Input takes a color in HEX. Best case a color we already have! 
Second takes percentage either as string or as a number 

`background: ${colorPercentage(colors.blueDark2, 20)}`


locothedev:
Thanks @BillyG83 <3

PR #20, opened 2022-01-10
feat(Card): Card component

locothedev:
Card component.
Customizable:
-tooltiptext
-title
-description
-content


src/components/Card/types.ts
```diff
@@ -0,0 +1,32 @@
+export interface CardProps {
+    
+    /**
+     * id will be generated automatically if not set
```
BillyG83, line 4: copy and paste strikes again =]

src/components/Card/Card.tsx
```diff
@@ -0,0 +1,70 @@
...
+                        ]
+                    }
+                    text={tooltipText}
+                />
+            </CardHeader>
+            <div>
+                {children}
+            </div>
+            <CardFooter>
+                <p data-testid={"title-test-id"}>{title}</p>
```
BillyG83, line 64: title is optional with no default, wont this render an empty <p> tag if the user forgets a title?

src/components/Card/Card.tsx
```diff
@@ -0,0 +1,70 @@
...
+            <CardHeader data-testid={"header-test-id"}>
+                { showCheckedIcon &&
+                    <Icon
+                    data-testid={"check-test-id"}
+                    fill={colors.green}
+                    size={24}
+                    svg={iconTypes.checkmark}
+                    />
+                }
+                <Tooltip
```
BillyG83, line 44: maybe tooltip should not render if there is no tooltipText prop passed, rather than rendering Lipsum. Or make it not optional if you need one

src/components/Card/Card.tsx
```diff
@@ -0,0 +1,70 @@
...
+                    data-testid={"tooltip-test-id"}
+                    children={
+                        [
+                            <Icon
+                                key={"ttip-card"}
```
BillyG83, line 49: was the console complaining about this needing a key? if its in a loop then it would need to be dynamic... but i dont think so in this case. hmmm i'm confused

also just FYI you need need to pass strings as JS
```key={"sting"} === key="string"```

**Review by BillyG83 — APPROVED** [review]
some minor feedback, merge when you want

File: src/components/Card/Card.tsx (hunk 4 above)
Comment (locothedev) [reply src/components/Card/Card.tsx:49]: Yup i need to pass in a key. Otherwise tests cry around

File: src/components/Card/Card.styles.ts [hunk 5]
```diff
@@ -0,0 +1,53 @@
...
+  color: ${colors.blue};
+  display: grid;
+  text-align: center;
+  & > p {
+    ${fonts.textBold}
+    margin: 0;
+  }
+`
+
+const selected = `
```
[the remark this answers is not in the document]
locothedev: Gotchya!

src/components/Card/Card.tsx
locothedev: fixed

src/components/Card/Card.tsx
locothedev: made it not optional

locothedev:
Made changes, commited and merging now

PR #21, opened 2022-01-10
translucent button

dani69654:


src/components/Button/Button.stories.tsx
```diff
@@ -391,3 +391,15 @@ DisabledIconOnlySmall.args = {
     iconLayout: 'icon-only',
     size: 'small',
 };
+
+export const Translucent = Template.bind({});
+Translucent.args = {
+    disabled: false,
```
BillyG83, line 397: please do not include props that are not needed. I'm pretty sure `disabled` and `type` are not needed in this case. This sounds like i am being annoying but when the next dev opens the code sample in storybook they will think they also need these props to make this button render, but i believe that is not the case sorry

BillyG83 (approved):
just remove the unneeded props, great job TY!

PR #28, opened 2022-01-11
Feat/input icons

Y0moo:
Customizable Input component added


src/components/Input/types.ts
```diff
@@ -42,5 +47,43 @@ export interface InputProps {
     /**
      * the input can use state to react to user interaction
      */
-    state?: 'error' | 'confirmed';
+    state?: 'error' | 'confirmed' | 'disabled';
```
BillyG83, line 50: I like this because we have state, but also `disabled` is a property that inputs use. So I wonder should we just use the standard HTML way like with Button? What do you think? Josef was keen on keeping as much props as standard when we started, like the onClick and onChange events too are standard HTML props

Y0moo: I started by adding the independent prop `disabled`. So it was possible to use it like this: <Input ... disabled />. But since we have the `state` prop which will be 100% used for doing custom logic like:
```
let inputState;
if (!hasAccess) inputState = "disabled"
if (dataIsValid) inputState = "confirmed"
<Input ... state={inputState} />
```

If there will be separated props and the user write something like `<Input error={true} confirmed={true} disabled={true} />`  we should work on logic like it's not allowed for example to have `confirmed` and `error` props `true`

Y0moo: or we can leave the state and add `disabled` only as independent

Y0moo: If the second option - I like it

BillyG83: i just think its slightly better DX to use the props the dev will expect rather than hiding it in state. But if you want to use state for logic that cool.
`<Input disbaled />`
to be is more intuitive than
z<Input state="disabled" />`

Y0moo: <Input disabled /> is amazing

Y0moo:
found a small bug with changing color of icon on hover. fixing

src/components/Input/types.ts
```diff
@@ -42,5 +47,48 @@ export interface InputProps {
...
+
+    /**
+     * hides the input text when is true
+     */
+    inputHidden?: boolean;
+
+    /**
+     * shows the hide icon when is true
+     */
+    hidable?: boolean;
```
locothedev, line 65: @Y0moo @BillyG83 
is it hideable or hidable. Im not sure

PR #29, opened 2022-01-11
feat/avatar

dani69654:


src/components/Avatar/Avatar.stories.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+export const AvatarImageDefault = Template.bind({});
+AvatarImageDefault.args = {
+    theme: 'image',
+};
+
+export const AvatarImageCustom = Template.bind({});
+AvatarImageCustom.args = {
+    theme: 'image',
+    image: 'https://scalebranding.com/wp-content/uploads/2021/06/penguin.jpg',
```
BillyG83, line 20: maybe we should point this to an image we host?

dani69654: Adding an image inside the component `Avatar/images` and using that one

BillyG83 (approved):
just the url comment in case the WP sites dies or removes it

src/components/Avatar/Avatar.stories.tsx
dani69654: Amended as suggested.

PR #33, opened 2022-01-12
Feat accordion

BillyG83:
- added TS comp
- used styled components
- set out nice Storybook layout
- tested all functional stories


BillyG83 (commented):
yep i think i'm happy but ive been coding for 13 hours so it could all be shit 😅

PR #34, opened 2022-01-12
chore: add `eslint`

oscario2:


BillyG83 (approved):
another job well done, TY for the carry @oscario2

PR #46, opened 2022-01-16
Feature/rounded avatar

rayyan224:


src/components/Avatar/Avatar.stories.tsx
```diff
@@ -25,3 +25,23 @@ AvatarLetters.args = {
     theme: 'letters',
     text: 'DM',
 };
+
+export const RoundedAvatarImageDefault = Template.bind({});
+RoundedAvatarImageDefault.args = {
+    theme: 'image',
+    rounded: true,
```
BillyG83, line 32: for a boolean please consider using a question.
EG: `isRounded` or `hasRounded`

we have tried to use this standard in the kit but if you see any we missed please point them out or better yet fix them please =]

src/components/Avatar/Avatar.tsx
```diff
@@ -3,9 +3,18 @@ import { AvatarProps } from './types';
 import React from 'react';
 import guy from './images/guy';
 
-const Avatar: React.FC<AvatarProps> = ({ theme, image, text }: AvatarProps) => {
+const Avatar: React.FC<AvatarProps> = ({
+    theme,
```
BillyG83, line 7: please make the props in alphabetical order. there is VSCode plugin called `sortLines`

src/components/Avatar/Avatar.tsx
```diff
@@ -3,9 +3,18 @@ import { AvatarProps } from './types';
...
+    image,
+    text,
+    rounded = false,
+}: AvatarProps) => {
     return (
-        <AvatarStyled data-testid="test-avatar">
+        <AvatarStyled
+            data-testid="test-avatar"
+            rounded={rounded}
+            role={`rounded_${rounded}`}
```
BillyG83, line 16: aria role tag is to provide semantic meaning to an impaired user. Role could be a nice tag to use here, but I would think that `role="Your Avatar"` would provide more meaning to a blind user. What do you think?

BillyG83 (changes requested):
minor feedback about rounded prop and role attribute, otherwise it looks perfect to me. Great job

PR #48, opened 2022-01-17
feat(Input): validation, better Stories, better tests

BillyG83:
- added validation on inputs when user clicks away
- cleaned up Storybook to reduce the number of stories
- clean up the tests to reduce them
- added validation tests
- changed markup to new standard agreements


locothedev:
PAMP

PR #51, opened 2022-01-18
Running prettier on commits

Y0moo:
- now `yarn prettier`  runs automatically on commits
- added .json to list of prettier controlled files


BillyG83 (approved):
beautiful!

PR #52, opened 2022-01-18
feat: add dropdown element component

dani69654:
This is the component that can/should be used to use when creating a dropdown menu.
The only goal of this component is to be 100% customisable, it also allows to set `onClick` event to run a function when clicked.

Height and width are auto if not specified.
Text, icon, colours, size, everything customisable.

### DIFF
--- a/src/components/Dropdown/Dropdown.stories.tsx
+++ b/src/components/Dropdown/Dropdown.stories.tsx
@@ -32,6 +32,7 @@ DropdownSelection.args = {
     position: 'bottom',
     children: [
         <DropdownElement
+            key="0"
             height={50}
             width={260}
             text={'Testnet Server'}
@@ -43,6 +44,7 @@ DropdownSelection.args = {
             onClick={() => alert('Testnet Server')}
         />,
         <DropdownElement
+            key="1"
             height={50}
             width={260}
             text={'Mainnet Server'}
@@ -54,6 +56,7 @@ DropdownSelection.args = {
             onClick={() => alert('Mainnet Server')}
         />,
         <DropdownElement
+            key="2"
             height={50}
             width={260}
             text={'Local Devchain Server'}
@@ -66,7 +69,7 @@ DropdownSelection.args = {
         />,
     ],
     parent: (
-        <Icon key="10" svg={iconTypes.helpCircle} fill={color.grey} size={50} />
+        <Icon key="3" svg={iconTypes.helpCircle} fill={color.grey} size={50} />
     ),
 };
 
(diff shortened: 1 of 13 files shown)


src/components/DropdownElement/DropdownElement.styles.tsx
```diff
@@ -0,0 +1,45 @@
...
+    width: ${(p) => (p.width ? `${p.width}px` : 'auto')};
+`;
+
+const DivImageStyled = styled.div`
+    align-items: center;
+    display: flex;
+    padding-right: 8px;
+`;
+
+const DivTextStyled = styled.p<
```
BillyG83, line 29: you can drop the `Div` here please, just `TextStyled` is perfect =]

src/components/DropdownElement/DropdownElement.tsx
```diff
@@ -0,0 +1,61 @@
...
+const { DivContainerStyled, DivStyled, DivImageStyled, DivTextStyled } =
+    DropdownElementStyles;
+
+const DropdownElement: React.FC<DropdownElementProps> = ({
+    backgroundColor = colors.blueDark,
+    height,
+    icon = iconTypes.check,
+    iconColor = 'white',
+    iconSize = 20,
+    id = String(Date.now()),
```
BillyG83, line 17: this ID thing was my idea, it worked in test but if there element renders super fast it can cause warnings, lets remove this please

src/components/DropdownElement/types.ts
```diff
@@ -0,0 +1,60 @@
...
+     */
+    iconColor?: string;
+
+    /**
+     * Icon size
+     */
+    iconSize?: number;
+
+    /**
+     * The tooltip ID will be generated if not assigned
```
BillyG83, line 30: update this comment when the ID comment is fixed please

src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -0,0 +1,115 @@
...
+            height={50}
+            width={260}
+            text={'Local Devchain Server'}
+            textSize={20}
+            iconSize={30}
+            backgroundColor={'transparent'}
+            icon={iconTypes.server}
+            textColor={colors.white}
+            onClick={() => alert('Local Devchain Server')}
+        />,
```
locothedev, line 66: add a key prop for every dropdown element in every story

src/components/DropdownElement/DropdownElement.tsx
dani69654: removed. Maybe we could use `uuid` for all the Components (I think that's the standard way to go for react apps)

PR #65, opened 2022-01-19
fix(Checkbox,Radios): validation types naming convention (fixes #64)

### LINKED ISSUE #64: Validation types naming convention
First of all I am a **fan of good naming conventions** and `web3uikit` is a great example. however I spotted a tiny naming issue as below:

The kit uses `RadiosValidate` and `CheckboxValidate` but Input use `ValidateInput`. I assume one post or prefix type name should be used. So which option do you prefer below?

1. Rename `ValidateInput` -> `InputValidate`
2. Rename `RadiosValidate` and `CheckboxValidate` -> `ValidateRadios` and `ValidateCheckbox`
3. Not important, leave as it is now :)

If you prefer 1. or 2. I can do a first PR as a warmup :)

Thanks for considering my minor naming issue.

attilacsanyi:
Fixing issue https://github.com/web3ui/web3uikit/issues/64

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 .DS_Store
 node_modules
 .cache
+.history
 dist
 /src/*/*/__snapshots__/
 /storybook-static/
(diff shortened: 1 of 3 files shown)


BillyG83 (approved):
good catch @attilacsanyi TY, merged!

attilacsanyi:
> good catch @attilacsanyi TY, merged!

Thanks :)

PR #66, opened 2022-01-20
Bugfix/form

rayyan224:


src/components/Button/types.ts
```diff
@@ -51,4 +51,10 @@ export interface ButtonProps {
      * set an icon position, or maybe show only the icon
      */
     iconLayout?: 'leading' | 'trailing' | 'icon-only';
+
+    /**
+     * Apply custom styles to button
+     * Will override all base styles
+     */
+    buttonStyle?: React.CSSProperties;
```
locothedev, line 59: I am not sure if @BillyG83 is a fan of too non specific desgin breaking styles ^^

locothedev:
@BillyG83 fixed that. Thanksss

PR #72, opened 2022-01-20
refactoring: styling+router link added

Y0moo:
<Breadcrumbs> component:

- Refactored the styling
- Integrated the react-router-dom directly to the breadcrumbs
- Added a router wrapper for correct displaying in stories


BillyG83 (approved):
very nice @Y0moo

PR #79, opened 2022-01-21
Feat/loader + Table Fix

rayyan224:


BillyG83 (approved):
great job @rayyan224 i'll think refactor the colors to use the colors interface when it is ready. Also remember to sort props alphabetically in CSS ad JS please, that just be being picky lol

PR #80, opened 2022-01-21
upd: Modal

Y0moo:
Added:
- `position: fixed` for displaying the Modal component correctly (see bug bellow)
- background blur for better UX
- interactive Modal Story

Changes:
- Instead of using the Stories for testing I've used `<Modal />` component because we cannot use Storybook preview hooks (`useArgs()`) outside `<Story>`. I've provided all `args` directly to the component

bug:
image


BillyG83 (approved):
great job @Y0moo i love the new args trick and the testing is cool too, nice one dude, great contribute on this PR

PR #82, opened 2022-01-21
Moralis Dark and Light theme / Storybook dark and light theme.

ChadChelios:
Added a Moralis color themed Light and Dark mode for the Storybook.

The default Storybook dark and light theme names are 'theme.dark' and 'theme.normal' if you prefer. 

**To use the default coloring, remove the 'import { create } from '@storybook/theming';' , 'const moralislightTheme'  and  'const moralislightTheme' and in .storybook/preview.js.

Some of the things that were changed were updates to my local environment using the following:

1) 'yarn upgrade' to update installed packages. 

2) Installing Storybooks "Addon Ally" (yarn add -D @storybook/addon-a11y) (https://storybook.js.org/docs/react/addons/install-addons/) (This can be replaced by installing only the Theme pluging (yarn add @react-theming/storybook-addon)

3) Adding the Storybook eslint plugin. 

I know we aren't supposed to make so many changes at once, but I wasn't sure how to disclude the local updates in this pull request. 

Let me know I need to change anything on my end for this pull request. I tried to notate as I went, but let me know if there is any gaps I can fill in.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
     "devDependencies": {
         "@babel/core": "^7.16.5",
         "@size-limit/preset-small-lib": "^7.0.5",
+        "@storybook/addon-a11y": "^6.4.14",
         "@storybook/addon-actions": "^6.4.9",
         "@storybook/addon-essentials": "^6.4.9",
         "@storybook/addon-info": "^5.3.21",
@@ -74,6 +75,7 @@
         "cssjson": "^2.1.3",
         "eslint": "^8.6.0",
         "eslint-config-google": "^0.14.0",
+        "eslint-plugin-storybook": "^0.5.6",
         "husky": "^7.0.4",
         "jest-styled-components": "^7.0.8",
         "jest-svg-transformer": "^1.0.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -83,9 +85,13 @@
...
         "rollup-plugin-inline-svg": "^2.0.0",
         "size-limit": "^7.0.5",
+        "storybook-dark-mode": "^1.0.8",
         "styled-components": "^5.3.3",
         "tsdx": "^0.14.1",
         "tslib": "^2.3.1",
         "typescript": "^4.5.4"
+    },
+    "dependencies": {
+        "web3uikit": "^0.1.16"
```
BillyG83, line 95: Hi dude, why would the kit be a dependency of itself?

BillyG83 (changes requested):
I left just one comment on the package.json please my friend, TY
otherwise this is cool good job

package.json
ChadChelios: Good question, I will remove. Good catch!

locothedev:
I do not think we need this feature now as it's adding one more dependency to the repo and is not really needed.

Thank you

ChadChelios:
> I do not think we need this feature now as it's adding one more dependency to the repo and is not really needed.
> 
> Thank you 

Oh boy, I was working on cleaning up the addon all night. Thought this might be the case since no one cared about it. Thanks for the opportunity, Moralis.

locothedev:
@ChadChelios  We will get back to this PR as soon we have themes like darkMode for all components. 

Thank you

ChadChelios:
I hope it works out well for you guys.

PR #85, opened 2022-01-22
Feat(colors) Added new colors interface

MarlenAw:
- [X] Added colors interface

* The changes that were made to the file **Accordion.tsx** were pushed by mistake. Changes for (Accordion.tsx) will be discarded once PR is approved. (For the meantime you can look at this as an example on how to use the interface in other components)


locothedev:
This is a wrong approach, in Accordion you are writing down the color-codes again which is redundant what leads to bad code . 
Thank you

PR #89, opened 2022-01-24
Bugfix/table and tabs fix

rayyan224:


BillyG83 (approved):
TY @rayyan224 so the speedy fix

PR #93, opened 2022-01-24
Feat/credit card

BillyG83:
- Radios had new feature added so they can be checked by default
- Radios had new feature added so user can pick which is checked, first by default
- Credit card updated to user radios new checked feature
- a few minor CSS UI issues were fixed
- markup was aligned


locothedev (changes requested):
We should have Mastercard and visa logo inside Logos folder already

PR #101, opened 2022-01-25
fix(Modal): hasCancel boolean added, onCancelButtonPress event added

BillyG83:
- hasCancel optional boolean added, default true, so cancel button can be hidden, CSS fix so ok is always aligned right even without a sibling
- remove all state React code, it should not govern its own state
- added onCancelButtonPress event so the next dev can flip isVisable or better yet unrender the Modal (docs added)
- minor alignment


locothedev:
Thank you bro

PR #108, opened 2022-01-26
feat-icons-addedCryptoIcons

johnvsnagendra:
Add some most used crypto coins icons.
Every crypto icon that are added are equally customizable as the normal icons and also comes with their original color when no color is selected.


locothedev:
@johnvsnagendra We use our own designed crypto icons inside Illustrations folder.
If you want you can use the svgs in illustration and modify them to be more customizable otherwise we have redundant svg's

johnvsnagendra:
> @johnvsnagendra We use our own designed crypto icons inside Illustrations folder.
> If you want you can use the svgs in illustration and modify them to be more customizable otherwise we have redundant svg's

So...u meant to recreate moralis themed crypto illustrations instead of using standard crypto icons.
Am I right?

BillyG83 (approved):
@johnvsnagendra i think we might have some fine tuning to do,  they colors on Shib worry me a bit lol. But overlay this is a valuable contribution and i think you for it

johnvsnagendra:
> @johnvsnagendra i think we might have some fine tuning to do, they colors on Shib worry me a bit lol. But overlay this is a valuable contribution and i think you for it

Thank You

PR #111, opened 2022-01-26
Feat/diabled cards

rayyan224:


BillyG83 (approved):
good fix TY @rayyan224

PR #115, opened 2022-01-29
feat(Avatar): Accessibility, alignment, responsive custom image

BillyG83:
- increased accessibility
- aligned the code a little with the kit
- custom images are now fully responsive
- fixed tests

### DIFF
--- a/src/components/Avatar/Avatar.styles.tsx
+++ b/src/components/Avatar/Avatar.styles.tsx
@@ -1,12 +1,20 @@
-import styled from 'styled-components';
+import styled, { css } from 'styled-components';
 import fonts from '../../styles/fonts';
 import color from '../../styles/colors';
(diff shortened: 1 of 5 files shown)


src/components/Avatar/Avatar.styles.tsx
```diff
@@ -1,12 +1,20 @@
...
+
+export const roundedEdgeValue = '20px';
+
+const customBackgroundImage = css<TStyleProps>`
+    background-position: center;
+    background-size: cover;
+    background-image: url(${(p) => p.image});
+`;
+
+const DivStyled = styled.div<TStyleProps>`
```
Y0moo, line 17: I guess we need to make the `display: inline;` rule for the element because usually, we use text in the same block.

image

PR #128, opened 2022-02-04
linebreak-style .eslintrc.js

Y0moo:
Fix for eslint error when working on windows. Without this rule code for people on windows looks like this:

image

Since we have people working on different OS and code editors I think we should use this rule.

https://eslint.org/docs/rules/linebreak-style#windows


locothedev:
Dear Bill Gates, please fix windows es lint issues

PR #136, opened 2022-02-11
Feat/stepper

BillyG83:
- added centralised style elements for things like headings
- added test ID to Loading to test in Stepper
- added new stepper component
- updated and sorted index for project


src/components/Stepper/Stepper.styles.ts
```diff
@@ -0,0 +1,122 @@
...
+type TStyleProps = Pick<
+    StepNumberProps,
+    'activeStep' | 'thisStep' | 'stepTotal'
+>;
+
+export const SectionStyled = styled.section`
+    display: block;
+    position: relative;
+    height: 100%;
+    display: flex;
```
locothedev, line 16: display block is overwritten and obsolete

locothedev (changes requested):
❤️

PR #143, opened 2022-02-15
Feat/ConnectButton without Moralis Server

Y0moo:
The `<ConnectButton />` component allows you to [authenticate](https://github.com/MoralisWeb3/react-moralis#authenticate-web3) users in case your server is initialized. When the server is not initialized, for example, you have `<MoralisProvider initializeOnMount={false} >` and you don't want to connect your Moralis server to the frontend the smart component will call [enableWeb3()](https://github.com/MoralisWeb3/react-moralis#enable-web3-via-metamask)

If you want to use this component with the connected server but without adding a user to Moralis Database you can add `moralisAuth` prop `<ConnectButton moralisAuth={false} />`

The  ConnectButton component automatically adds to the local storage info about the connector you used and will automatically call `enableWeb3()`. So if users was connected once it will automatically initialize web3 connection(No need anymore to add UseEffect hook for `enableWeb3()`


Y0moo:
some bugs, fixing

locothedev:
greatest of all time

PR #144, opened 2022-02-15
feat(Checkbox): added switchable text

BillyG83:
- added switchable text prop `labelWhenChecked` needed for Admin
- added new test
- added logic need
- added Story


locothedev (changes requested):
can you please remove yarn.lock from your pr?

PR #145, opened 2022-02-15
feat(DatePicker): comp added, styled, tested

BillyG83:
- added React / TS component
- harnessed HTML5 for best UX, accessibility and dev speed
- styled using shared styles of Input
- tested
- storybook
- project index updated with `<DatePicker />`


src/components/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,110 @@
...
+    const hasValidation = () =>
+        Boolean(
+            validation?.required ||
+                validation?.max ||
+                validation?.min ||
+                validation?.required,
+        );
+
+    const validate = (event: React.FocusEvent<HTMLInputElement, Element>) => {
+        if (!hasValidation()) return;
```
locothedev, line 44: whats that second argument`? Seems its not being used

src/components/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,110 @@
...
+    name = 'Select a date',
+    onChange,
+    state = disabled ? 'disabled' : undefined,
+    type = 'date',
+    validation,
+    value = new Date().toISOString().substring(0, type === 'date' ? 10 : 7),
+}) => {
+    const [current, setCurrent] = useState(value);
+    const [currentState, setCurrentState] = useState(state);
+    const [invalidMessage, setInvalidMessage] = useState('');
```
locothedev, line 26: what about types for current state so we not have to call `setCurrentState(undefined)` ?

locothedev (commented):
Such a great component

PR #152, opened 2022-02-17
feat(Typography): copyable prop added

Y0moo:
`<Typography copyable>` allows to copy content of the component

### DIFF
--- a/src/components/Typography/Typography.tsx
+++ b/src/components/Typography/Typography.tsx
@@ -1,9 +1,8 @@
-import React, { useState } from 'react';
+import React from 'react';
 import styled from 'styled-components';
-import color from '../../styles/colors';
-import { Icon, iconTypes } from '../Icon';
+import { CopyButton } from '../CopyButton';
 import { TypographyProps, variantType } from './types';
-import { CopyIconStyled, getTypographyStyle } from './Typography.styles';
+import { getTypographyStyle } from './Typography.styles';
 
 const getTag = (variant: variantType) => {
     switch (variant) {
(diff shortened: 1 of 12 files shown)


src/components/Typography/Typography.tsx
```diff
@@ -29,15 +31,45 @@ const DynamicText = ({
...
+
+    const copyToClipboard = (): void => {
+        if (typeof navigator == 'undefined') return;
+        navigator.clipboard.writeText(`${children}`);
+        setIsCopied(true);
+    };
+
+    const renderCopyIcon = () =>
+        copyable && (
+            <CopyIconStyled
```
BillyG83, line 49: this is nearly perfect and a great PR... all my cooments are right on this line

1. the naming convention is not aligned here... a styled button should be `ButtonStyled` and if you had one already, this could be `ButtonStyledCopy`
2. We should really use <Button /> props: icon, small, transparent BUT i do see that the icon changes so....
3. Really this `<CopyButton />` needs to be its own compoents (its used in a few places in admin and now twice in our kit

BUT, lets not block the PR, can you please open a task to make `<CopyButton />` and replace it here and in `<Input />`

BillyG83: I made the issue 153 so you can disregard point 2 and 3, but could you take a quick pass at the naming of the styled elements in this component please?

https://github.com/web3ui/web3uikit/issues/153

locothedev:
@Y0moo dont forget to export the new components in `src/index.ts` 🤞

BillyG83:
@locothedev nice call on 
`dont forget to export the copybutton component in src/index.ts`

I will add it to #153

PR #154, opened 2022-02-17
Feat/tooltip/refactor

rayyan224:


src/components/Tooltip/Tooltip.tsx
```diff
@@ -1,24 +1,63 @@
...
+    const popoverRef = useRef(null);
+    const [width, setWidth] = useState(0);
+    const [height, setHeight] = useState(0);
+    const [popverWidth, setPopoverWidth] = useState(0);
+    const [popoverHeight, setPopoverHeight] = useState(0);
+    useEffect(() => {
+        console.log(
+            'width',
+            parretnRef.current ? (parretnRef.current as any).offsetWidth : 0,
+        );
```
locothedev, line 26: console.log no bueno

src/components/Tooltip/types.ts
```diff
@@ -22,7 +17,7 @@ export interface TooltipProps {
     position: Position;
 
     /**
-     * Set text to show in tooltip
+     * Set element/text that will be shown within the tooltip
      */
-    text?: string;
+    content?: JSX.Element | string;
```
locothedev, line 22: lets make content required? What do you think about it

src/components/Tooltip/Tooltip.tsx
```diff
@@ -1,24 +1,63 @@
...
-            position={position}
-            data-tooltip={text}
+        <StyledTooltipParentDiv
+            ref={parretnRef}
             data-testid={'tooltip-container-test-id'}
-            maxWidth={maxWidth}
         >
-            <div data-testid={'tooltip-children-test-id'}>{children}</div>
-        </TooltipContainer>
+            <span>{children}</span>
```
locothedev, line 45: as we are allowing to pass in jsx elements I dont think its good allowing to pass in `div` inside `span`.

What is the actual difference between children and content?

locothedev (changes requested):
Awesome bro 👍

src/components/Tooltip/types.ts
rayyan224: Yea, ur right

src/components/Tooltip/Tooltip.tsx
rayyan224: Children is for what is shown 
content is what's shown in the tooltip

PR #171, opened 2022-03-02
Feat/crypto logos

Y0moo:
- Added separated CryptoLogos components. 
- fixed and improved responsible positioning and sizing for the `<Illustration />` and `<CryptoCards />` components

This component will be used for `<ChainSwitcher />` in web3uikit and moralis-admin-panel


BillyG83 (approved):
i see you add styles to `<Illustrations/>`, as its just one style and there is no style file, we will let this one slip in ;)

PR #172, opened 2022-03-02
feat: added cenetering for loading, and custom loading props, and stories

rayyan224:
Fix for centering loading on buttons. Also added loading props to allow users to override specific styles if necessary such as Spinner color.


BillyG83 (approved):
Nice sorry for the mix up with Icon

PR #183, opened 2022-03-03
Feat/icon size copy text

rayyan224:
icon size copy text

### DIFF
--- a/src/components/Typography/Typography.tsx
+++ b/src/components/Typography/Typography.tsx
@@ -32,6 +32,7 @@ const DynamicText = ({
     monospace,
     copyable,
     children,
+    iconSize,
     onCopy = () => {},
     ...otherProps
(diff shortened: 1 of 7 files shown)


src/components/Typography/Typography.tsx
```diff
@@ -32,6 +32,7 @@ const DynamicText = ({
     monospace,
     copyable,
     children,
+    iconSize = 22,
```
Y0moo, line 35: default Icon size for CopyButton is 1em, which means it takes the max height of the parent element. Why do we hardcode here `22`?

rayyan224: Did'nt see the 1em, good catch bro

BillyG83 (approved):
ok nice

BillyG83 (approved):
Niiiiice guys 🤩

PR #191, opened 2022-03-04
Breaking Change: BannerStrip buttonConfig + height

locothedev:
BREAKING CHANGES:
replaced button props with buttonConfigProp to have more control over button within BannerStrip


BillyG83:
could you achieve the same thing by adding optional height? then we don't have a breaking change

rayyan224 (approved):
Nice!

PR #197, opened 2022-03-04
feat: just thinking ahead of possible ways this could have bad ux

rayyan224:
Giving a max width and height, so if a user just has one icon, by default it would span the whole, page. We don't want that 😅


BillyG83 (approved):
i can't imagine this will do much, but hold onto your butts xD

PR #234, opened 2022-03-08
Feat form more inputs 207

BillyG83:
- Removed Radio from CreditCard
- Made new CreditCard picker from Radio if formatted CC data is passed
- Form can use Radio CC picker and can build formatted data to be returned
- cleaned up docs of Stepper a little


src/components/Form/Form.tsx
```diff
@@ -35,38 +37,67 @@ const Form: React.FC<FormProps> = ({
...
         if (event.target.checked) {
-            optionAdded(index, opt, isRadio);
+            optionAdded(elementIndex, opt, isRadio);
         } else {
-            const filtered = data[index]?.selected?.filter(
+            const filtered = data[elementIndex]?.selected?.filter(
                 (item) => item !== opt,
             );
-            data[index].selected = filtered;
+            data[elementIndex].selected = filtered;
```
locothedev, line 76: can simply it by:
`data[elementIndex].selected = data[elementIndex]?.selected?.filter(
      (item) => item !== opt,
 );`

src/components/CreditCard/CreditCard.styles.ts
```diff
@@ -31,23 +31,24 @@ export const DivStyledCreditCard = styled.div<TStyleProps>`
...
     ${({ isExpired, pressed }) =>
         isExpired ? expiredStyles : pressed ? pressedStyles : ''}
 `;
 
 export const DivStyledFlex = styled.div`
-    display: flex;
-    justify-content: space-between;
     align-items: center;
+    display: flex;
+    justify-content: flex-end;
```
locothedev, line 41: This style effects the creditcards footer and therefore the style does not match anymore

Should look like:
creditCardFigma

But looks like:
looksLike

locothedev (changes requested):

src/components/Form/Form.tsx
BillyG83: Do you think having a one liner is more simple?
I agree its less code, but I added the extra lines to make it more simple for the next dev to understand.
What do you think?

src/components/CreditCard/CreditCard.styles.ts
BillyG83: ok nice, will fix now

PR #247, opened 2022-03-09
Feat/dropdown hidelabel

Y0moo:
added prop hidelabel
image

### DIFF
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -20,6 +20,7 @@ const Dropdown: React.FC<IDropdown> = ({
     icon,
     isDisabled = false,
     isLabelFixed = true,
+    isLabelVisible = true,
     label,
     onChange,
     options,
@@ -87,13 +88,15 @@ const Dropdown: React.FC<IDropdown> = ({
                             key={option.id}
                         >
                             {option?.prefix}
-                            <Typography
-                                variant="caption14"
-                                weight="600"
-                                color="#041836"
-                            >
-                                {option.label}
-                            </Typography>
+                            {isLabelVisible && (
+                                <Typography
+                                    variant="caption14"
+                                    weight="600"
+                                    color="#041836"
+                                >
+                                    {option.label}
+                                </Typography>
+                            )}
                         </DivStyledOptionItem>
                     ))}
             </DivStyledOptionsContainer>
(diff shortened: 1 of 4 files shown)


rayyan224 (approved):

rayyan224:
I approved, but I think we should have a story for this, its a unique prop

src/components/Dropdown/Dropdown.tsx
```diff
@@ -20,6 +20,7 @@ const Dropdown: React.FC<IDropdown> = ({
     icon,
     isDisabled = false,
     isLabelFixed = true,
+    isLabelVisible = true,
```
BillyG83, line 23: could `isLabelVisible` maybe `hasLabel`?

and could that be set dynamically by `label` being an optional prop? That would be killer DX

`const hasLabel = Boolean(label)` for example

Y0moo:
Thank you @rayyan224. I'm agree with you, will add ASAP

BillyG83 (changes requested):
what do you think about my comment? this is 100% sweet but i think we could avoid devs needing to think about 2 props

src/components/Dropdown/Dropdown.tsx
Y0moo: Thank you @BillyG83 
If we do a `label` is not mandatory, we even don't need to have `isLabelVisible` and `hasLabel`.

But it is called `isLabelVisible` and not `hasLabel` because it can have a label but in specific conditions, we may want to hide the label.

Y0moo: image

Y0moo: e.g. full width submenu and collapsed

Y0moo: I think we don't really need `isLabelFixed`, because we can make it conditional @rayyan224 what do you think?

Y0moo: Yes, of course we can just use two Dropdowns and two different arrays of options, but do you think `isLabelVisible` a bad solution?

rayyan224: Hmm by conditionally rendering do you mean having the label withing the text for the option item ?

rayyan224:
@Y0moo Might be out of this scope but it might be nice to have this go into this smaller view as soon as, the width of the item is under a certain width? You can use a ref to get the width of the item. It would give a nice developer experience

BillyG83:
@Y0moo you clearly need this feature and explained your reasons very clearly.

If we can improve later, great. For now i don't want to block your progress. TY both guys, great input

BillyG83:
wow i closed that PR by accident, MY BAD!
it merged now

PR #249, opened 2022-03-09
feat(Tag): buffed discount mode options

BillyG83:
- in discount mode width = height and 50% border radius = perfect circle
- added font size to we can make big ass text

### DIFF
--- a/src/components/Tag/Tag.styles.ts
+++ b/src/components/Tag/Tag.styles.ts
@@ -69,11 +69,9 @@ export const TagStyled = styled.div<TStyleProps>`
     width: ${({ width }) => width};
 
     ${({ width, theme }) =>
-        theme === 'discount' && Boolean(width) && `height: ${width}`};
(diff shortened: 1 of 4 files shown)


rayyan224 (approved):

src/components/Tag/Tag.styles.ts
```diff
@@ -64,4 +67,13 @@ export const TagStyled = styled.div<TStyleProps>`
     ${({ color, theme, tone }) =>
         theme !== 'status' && color && getColors(color, tone)}
     width: ${({ width }) => width};
+
+    ${({ width, theme }) =>
```
rayyan224, line 71: if they have the same logic y not combine. It would be easier on the CSS in this case, a case so forex
`height: ${width}; width: ${width}; border-radius: 50%;`. It's less heavy on the browser this way

BillyG83: good call dude, fixed

Issue #255, opened 2022-03-10
Modal - Missing title will move the close button to top left corner 


dani69654:
Without `title` prop:
<img width="1023" alt="Screenshot 2022-03-10 at 09 29 25" src="https://user-images.githubusercontent.com/60792849/157620052-ee7f6b4e-170e-480f-9d0c-55ccf4375b1b.png">

With title:

<img width="1039" alt="Screenshot 2022-03-10 at 09 29 42" src="https://user-images.githubusercontent.com/60792849/157620105-385e5fc0-ca38-45c4-8947-13286197f2c4.png">

This might be related to the `justify-content: space-between;` prop in css

```js
export const HeaderStyled = styled.header`
    display: flex;
    align-items: center;
    justify-content: space-between; << ------
    padding: 24px 32px 10px;

    div {
        border-color: ${colors.blue};
        border-radius: 15px;
    }

    h3 {
        color: ${colors.blue};
        padding-right: 8px;
    }

    button {
        min-width: 30px;
    }
`;
```

PR #256, opened 2022-03-10
fix: modal button position when title is undefined

dani69654:
With title:
Screenshot 2022-03-10 at 09 51 44

Without title:
Screenshot 2022-03-10 at 09 51 59


dani69654:
Closes #255

Issue #284, opened 2022-03-15
[Bug] Dropdown menu tilted right side


dani69654:
<img width="1198" alt="Screenshot 2022-03-15 at 09 32 44" src="https://user-images.githubusercontent.com/60792849/158337553-91360852-13cd-4849-b780-7b75c1570fa1.png">

dani69654:
<img width="302" alt="Screenshot 2022-03-15 at 10 47 51" src="https://user-images.githubusercontent.com/60792849/158351318-c321c609-4964-4144-8bba-c79aa2277bd2.png">

PR #285, opened 2022-03-15
fix: dropdown item overflow and font weight

dani69654:


dani69654:
https://github.com/web3ui/web3uikit/issues/284

PR #297, opened 2022-03-18
fix: nested close buttons Modal component

dani69654:
Fix warning nested div in Modal component. <br />

Screenshot 2022-03-18 at 01 56 04


dani69654:
@locothedev 
This PR follows a great idea that @BillyG83 had. <br />
This idea was implemented first in the `Stepper` component, and consists is assigning an event listener to a custom `Button`.
Whenever a specific trigger is fired (`click` in this case) we check which element_id fired it, and we run a function in case the `id` matches the one we expect.

locothedev (approved):
remove console.log as discussed and merge after it

PR #300, opened 2022-03-18
feat: custom overflow rule Modal component

dani69654:
Alt Text

### DIFF
--- a/src/components/Modal/Modal.styles.ts
+++ b/src/components/Modal/Modal.styles.ts
@@ -3,8 +3,24 @@ import colors from '../../styles/colors';
 import fonts from '../../styles/fonts';
 import { ModalProps } from './types';
 
-type TStyleProps = Pick<ModalProps, 'isVisible' | 'hasCancel' | 'width'>;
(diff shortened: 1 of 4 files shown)


dani69654:
Added a `canOverflow` (bool) prop to add custom overflow rule if needed. <br />
Devs can now choose if the content of the modal should overflow (useful in case of dropdown menus)

src/components/Modal/Modal.styles.ts
```diff
@@ -3,8 +3,24 @@ import colors from '../../styles/colors';
...
+        ? ''
+        : `
+                overflow: auto;
+                ::-webkit-scrollbar {
+                    display: none;
+                }
+                scrollbar-width: none;
+                -ms-overflow-style: none;
+            `;
+};
```
locothedev, line 23: console.log

maybe its better to return the css after an if statement so we dont return an empty string;

`if(!v) return css`

locothedev (approved):
check minor comment

locothedev (approved):
nooiccee

PR #301, opened 2022-03-18
feat: allow ReactNode element as final message in Stepper component

dani69654:
Alt Text


dani69654:
As requested by designer, we should have a way to render a dom element in addition to the `string` type we already allow.

locothedev (approved):
clean

PR #319, opened 2022-03-22
fix(font): color + weight

locothedev:


locothedev:
match new design. 

Bug found by @MateusbrMedeiros

PR #321, opened 2022-03-22
added the props:headerHasBottomBorder to the modal component

akinolu52:
This PR will add bottom border to header of a modal component
the flag name is: headerHasBottomBorder
type is: boolean
default value: false

I also added an example to the story file to enable quick access to the example(demo)

Screenshot 2022-03-22 at 15 13 25


locothedev (approved):
clean!

PR #327, opened 2022-03-23
fix(Input): removing bad value fix

BillyG83:
i have shame 😓


locothedev (changes requested):
we will need to refactor a lot here. Behaviour on rerender no bueno.

Need to discuss before opening up an Issue

BillyG83:
I agree with you @locothedev we can improve this component if you want, but that was a bug that needs removed first

PR #337, opened 2022-03-28
fix: update TokenPocket wallet logo

chendatony31:


src/components/WalletModal/WalletIcons/TokenPocket.tsx
```diff
@@ -2,139 +2,31 @@ import React from 'react';
...
     return (
-        <svg
-            xmlns="http://www.w3.org/2000/svg"
-            width="40"
-            height="40"
-            version="1.1"
-            viewBox="0 0 264.583 264.583"
-        >
+        <svg width="40" height="40" viewBox="0 0 1024 1024" fill="none" xmlns="http://www.w3.org/2000/svg">
+            <mask id="mask0_408_225" style="mask-type:alpha" maskUnits="userSpaceOnUse" x="0" y="0" width="1024" height="1024">
```
locothedev, line 6: string type in `style` is not allowed. 

Is it needed?

locothedev (changes requested):

src/components/WalletModal/WalletIcons/TokenPocket.tsx
chendatony31: Just deleted the `mask` element. I think all of it is not needed.

locothedev:
@chendatony31 sorry I have not pushed my comment. 

You need to replace every prop that has a `-` divider.
`stop-color` to `stopColor`
`fill-opacity` to `fillOpacity`
`clip-path` to `clipPath`
and any other occurance

PR #344, opened 2022-03-31
feat(React18): upgrade

locothedev:
Upgrade peerDep and DevDep to react 18

For anyone who is interested:

this PR needs some refactor on every test file as ReactDOM.render doesnt work anymore in react 18
If you want to use web3uikit in react 18 you probably see issues when trying to install with `npm `so please use `yarn` and everything works fine


akinolu52:
hi @locothedev 

I just checked react documentation, and it looks like a simple fix is available

Screenshot 2022-04-04 at 21 33 25

locothedev:
> hi @locothedev
> 
> I just checked react documentation, and it looks like a simple fix is available
> 
> Screenshot 2022-04-04 at 21 33 25

Yeah I know that its quite simple to refactor but refactoring around 40 tests files is time consuming

locothedev:
BEFORE REFACTOR `import ReactDOM from 'react-dom';`
before

AFTER REFACTOR TO `import ReactDOM from 'react-dom/client';`
after

locothedev:
> BEFORE REFACTOR `import ReactDOM from 'react-dom';` before
> 
> AFTER REFACTOR TO `import ReactDOM from 'react-dom/client';` after

we either refactor that way or use `render` from RTL@13-alpha. Still both ways are time consuming. 
40 files and some of them have multiple test blocks so it could be more than 100 block to refactor

SmithVyne:
> > BEFORE REFACTOR `import ReactDOM from 'react-dom';` before
> > AFTER REFACTOR TO `import ReactDOM from 'react-dom/client';` after
> 
> we either refactor that way or use `render` from RTL@13-alpha. Still both ways are time consuming. 40 files and some of them have multiple test blocks so it could be more than 100 blocks to refactor

Personally, I like the render from RTL more, but are there any benefits of using render in RTL over createRoot?

locothedev:
> > > BEFORE REFACTOR `import ReactDOM from 'react-dom';` before
> > > AFTER REFACTOR TO `import ReactDOM from 'react-dom/client';` after
> > 
> > 
> > we either refactor that way or use `render` from RTL@13-alpha. Still both ways are time consuming. 40 files and some of them have multiple test blocks so it could be more than 100 blocks to refactor
> 
> Personally, I like the render from RTL more, but are there any benefits of using createRoot over render in RTL? Also RTL is a lot cleaner

I totally agree on this - I am not sure how much more we need to refactor but I know its a lot actually. 
Also why I agree on this in the admin project we are also using render from rtl@13-alpha

SmithVyne:
> > > > BEFORE REFACTOR `import ReactDOM from 'react-dom';` before
> 
> > > > AFTER REFACTOR TO `import ReactDOM from 'react-dom/client';` after
> 
> > > 
> 
> > > 
> 
> > > we either refactor that way or use `render` from RTL@13-alpha. Still both ways are time consuming. 40 files and some of them have multiple test blocks so it could be more than 100 blocks to refactor
> 
> > 
> 
> > Personally, I like the render from RTL more, but are there any benefits of using createRoot over render in RTL? Also RTL is a lot cleaner
> 
> 
> 
> I totally agree on this - I am not sure how much more we need to refactor but I know its a lot actually. 
> 
> Also why I agree on this in the admin project we are also using render from rtl@13-alpha

Yeah, it'll be a lot of refactoring but we also we have no choice but to upgrade React so we have to do it.

locothedev:
I personally hope for @oscario2 to have a script that does all that for us lol 🤣

locothedev:
OK I went through the tests some of them already use rtl's render. 

I found 156 results in 34 files.

BillyG83:
yeah Rayyan did his own way thankfully @locothedev can we open a task to fix one that small to see how much effort is involved?

locothedev:
> yeah Rayyan did his own way thankfully @locothedev can we open a task to fix one that small to see how much effort is involved?

not that much effort. I think 2 people is enough for this task and could be finished fast

SmithVyne:
> > yeah Rayyan did his own way thankfully @locothedev can we open a task to fix one that small to see how much effort is involved?
> 
> not that much effort. I think 2 people is enough for this task and could be finished fast

I'm looking at the code to see what this looks like. Most changes should look something like this:
image.
Although there's more work to be done for the beforeAll's, afterAll's, etc

BillyG83:
this task is blocked by all tasks marked `React18`

PR #349, opened 2022-04-04
Fix readme incorrect links

osamajandali:
- Fixed all stories links
- I used only the component page without the properties so the link will always redirect to the default state


BillyG83 (approved):
great job @osamajandali TY so much

PR #353, opened 2022-04-05
feature: chain selector

dani69654:
Screenshot 2022-04-05 at 13 15 13


Y0moo:
Added state to display selected chains in the storybook demo

PR #394, opened 2022-04-06
add signingMessage to ConnectButton

akinolu52:
added the signingMessage props, so as to enable custom message to the <Connectbutton />

### DIFF
--- a/src/components/ConnectButton/ConnectButton.tsx
+++ b/src/components/ConnectButton/ConnectButton.tsx
@@ -21,6 +21,7 @@ type web3StatusType = 'disconnected' | 'pending' | 'only_web3';
 
 const ConnectButton: React.FC<ConnectButtonProps> = ({
     moralisAuth = true,
+    signingMessage = undefined,
 }) => {
     const {
         account,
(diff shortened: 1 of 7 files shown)


src/components/ConnectButton/ConnectButton.tsx
```diff
@@ -20,7 +20,9 @@ const {
 type web3StatusType = 'disconnected' | 'pending' | 'only_web3';
 
 const ConnectButton: React.FC<ConnectButtonProps> = ({
+    chainId = undefined,
```
BillyG83, line 23: Do you need to assign `undefined`?

I believe it is the same thing to have
```
chainId,
moralisAuth = true,
signingMessage,
```

BillyG83 (commented):
looks great but one question for you on ConnectButton if you dont mind, TY @akinolu52

Y0moo (commented):
The thing is that authentication happens in WalletModal and not in connect button.

BillyG83:
TY @Y0moo did you get that feedback ok @akinolu52 ?

https://github.com/web3ui/web3uikit/pull/394#pullrequestreview-933319553

src/components/ConnectButton/ConnectButton.tsx
akinolu52: it's removed (key and value) using `JSON.stringify`
if I dont use undefined i believed it'll be passed as an empty string to the backend, and i'm not sure if an empty string is being handled on the server. @BillyG83

akinolu52:
Fixed in the latest update @Y0moo

Y0moo (approved):

src/components/WalletModal/types.ts
```diff
@@ -1,5 +1,10 @@
 import { ConnectButtonProps } from '../ConnectButton/types';
 export interface WalletModalProps extends ConnectButtonProps {
+    /**
+     * an optional chain id of the blockchain that the web3 wallet is connected to
+     */
+    chainId?: number;
```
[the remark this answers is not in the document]
Y0moo: checked, in this case, we need to sue number

Issue #271, opened 2022-03-11
[Feat]: Input component to also allow `Blockie` instead of an icon


dani69654:
Implementing the following as per designer req.
![Screenshot 2022-03-11 at 15 25 52](https://user-images.githubusercontent.com/60792849/157886249-9804689f-f173-4193-b89a-0961a39eb2ac.png)

AbhinavMV:
Hey @osamajandali 
instead of writing entirely new code maybe we can use Blockie component.
In Button> type.tsx
```tsx
prefixIcon?: TIconType | 'blockie';
blockieSeed?: string;
```
And then in Button.tsx
```tsx
<DivStyled className="input_prefixIcon">
      {prefixIcon !== 'blockie' && <Icon svg={prefixIcon} />}
      {prefixIcon === 'blockie' && <Blockie seed={blockieSeed} />}
</DivStyled>
```
Let me know if you need any help 😄

osamajandali:
@AbhinavMV I already did that but the problem isn’t here in the blockie the problem was with the text being ellipsed on blur
![image](https://user-images.githubusercontent.com/13832392/166105377-7604a6fe-fe47-4fb8-93aa-9c7aa62be3ff.png)

and you can check here it's done #396, need some code refactor only

PR #396, opened 2022-04-06
[feat] Wallet address input

osamajandali:
Reference : Issue #271 

Add wallet address input type which includes:
- Text ellipsis on blur
- Prefix blockie that changes with the input

Here is how it look like
wallet input

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,7 +92,6 @@
         "eslint": "^8.6.0",
         "eslint-config-google": "^0.14.0",
         "husky": "^7.0.4",
-        "jest-canvas-mock": "^2.3.1",
         "jest-styled-components": "^7.0.8",
         "jest-svg-transformer": "^1.0.0",
         "postcss": "^8.4.6",
--- a/package.json
+++ b/package.json
@@ -92,6 +92,7 @@
         "eslint": "^8.6.0",
         "eslint-config-google": "^0.14.0",
         "husky": "^7.0.4",
+        "jest-canvas-mock": "^2.3.1",
         "jest-styled-components": "^7.0.8",
         "jest-svg-transformer": "^1.0.0",
         "postcss": "^8.4.6",
--- a/src/components/Input/Input.stories.tsx
+++ b/src/components/Input/Input.stories.tsx
@@ -77,13 +77,6 @@ PasswordInput.args = {
     name: 'Test Password Input',
 };
 
-export const WalletInput = Template.bind({});
-WalletInput.args = {
-    label: 'Address',
-    name: 'Test address',
-    isWalletAddress: true,
-};
-
 export const PrefixCopyableHidableActive = Template.bind({});
 PrefixCopyableHidableActive.args = {
     hasCopyButton: true,
--- a/src/components/Input/Input.stories.tsx
+++ b/src/components/Input/Input.stories.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { ComponentStory, ComponentMeta } from '@storybook/react';
 import Input from '../../components/Input/Input';
+import WalletInputComponent from '../../components/Input/WalletInput';
 import { iconTypes } from '../Icon/collection';
 import { validateRegExp } from '../../utils/const';
 
@@ -19,6 +20,9 @@ const FilledTemplate: ComponentStory<typeof Input> = (args) => (
     <div className="input_filled">
         <Input {...args} />
     </div>
+);
+const WalletTemplate: ComponentStory<typeof Input> = (args) => (
+    <WalletInputComponent {...args} />
 );
 
 export const DefaultInput = Template.bind({});
@@ -75,6 +79,12 @@ PasswordInput.args = {
     type: 'password',
     label: 'enter a password',
     name: 'Test Password Input',
+};
+
+export const WalletInput = WalletTemplate.bind({});
+WalletInput.args = {
+    label: 'Address',
+    name: 'Test address',
 };
 
 export const PrefixCopyableHidableActive = Template.bind({});
--- a/src/components/Input/Input.stories.tsx
+++ b/src/components/Input/Input.stories.tsx
@@ -77,6 +77,13 @@ PasswordInput.args = {
     name: 'Test Password Input',
 };
 
+export const WalletInput = Template.bind({});
+WalletInput.args = {
+    label: 'Address',
+    name: 'Test address',
+    isWalletAddress: true,
+};
+
 export const PrefixCopyableHidableActive = Template.bind({});
 PrefixCopyableHidableActive.args = {
     hasCopyButton: true,
--- a/src/components/Input/Input.styles.ts
+++ b/src/components/Input/Input.styles.ts
@@ -174,21 +174,6 @@ export const DivStyled = styled.div`
     margin-right: 12px;
 `;
 
-export const BlockieWrapperStyled = styled.div`
-    width: 24px;
-    height: 24px;
-    display: flex;
-    margin-right: 12px;
-    & .blockie {
-        border-radius: 50%;
-    }
-`;
-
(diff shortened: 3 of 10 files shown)


Mitumaru (approved):

package.json
```diff
@@ -92,6 +92,7 @@
         "eslint": "^8.6.0",
         "eslint-config-google": "^0.14.0",
         "husky": "^7.0.4",
+        "jest-canvas-mock": "^2.3.1",
```
BillyG83, line 95: what is this used for? I dont see it in your test

src/components/Input/Input.styles.ts
```diff
@@ -174,6 +174,21 @@ export const DivStyled = styled.div`
     margin-right: 12px;
 `;
 
+export const BlockieWrapperStyled = styled.div`
```
BillyG83, line 177: I think the naming convention might not have even existed back then but:
https://github.com/web3ui/web3uikit/blob/master/CONTRIBUTE.md#css-naming-an-element
Could you rename them `DivStyledBlockie` and `DivStyledEllipsis` please?

src/components/Input/Input.test.tsx
```diff
@@ -3,19 +3,22 @@ import ReactDOM from 'react-dom';
 import { render, screen, fireEvent } from '@testing-library/react';
 import '@testing-library/jest-dom';
 import 'jest-styled-components';
+import 'jest-canvas-mock';
```
BillyG83, line 6: i see you import it here but what do you use it for?

src/components/Input/Input.tsx
```diff
@@ -34,16 +38,27 @@ const Input: React.FC<InputProps> = ({
...
     const [currentState, setCurrentState] = useState(state);
     const [mainType, setMainType] = useState(type);
     const [isInputHidden, setIsInputHidden] = useState(inputHidden);
     const [invalidMessage, setInvalidMessage] = useState(errorMessage);
+    const [isEditMode, setIsEditMode] = useState(true);
 
     useEffect(() => setIsInputHidden(type === 'password'), [inputHidden]);
     useEffect(() => setCurrentState(state), [state]);
     useEffect(() => setMainType(type), [type]);
+    useEffect(() => {
```
BillyG83, line 57: this is cool and it helps with your input, but checkout the other stories, especially checkout Form... maybe it is worth considering to make a new component like you said, this one already has a lot of logic.

If the new component called in Input with type text and Blockie, would it not be easier to control their state from a parent without effecting their core functionality perhaps?

BillyG83 (changes requested):
you did a lot of great work here, the code is very standardised and your clearly read the docs. Sorry if some old naming crimes confused you. This component was maybe the 2nd or 3rd i made 😅 

But i think it is a better idea to build a new component. Check my comments and let me know what you think please. Again great work on this... Input is a tricky one to jump into

package.json
osamajandali: Needed to test the blockie because it is a canvas element 
Otherwise you won’t able to render the whole component in the tests

src/components/Input/Input.tsx
osamajandali: I think it is not doable with the current input structure as there is no shared functionality, because here we need to edit the input to look and behave in a certain way so it's not possible to add on top of it.

BillyG83: can you explain that more for me sorry 😅 

Input onChange will return the value, can you not store that string in a parent component and feed it to the Blockie compoent?

osamajandali: Thr blockie is inside the input and it should replace the prefix so how could I add blockie if I grap the normal Input
That will require changes to the Input field to accept components for the prefix so it could then accept blockie and put it there

That previous isn't the problem, the problem occurs when you want to display ellipsis text I have then to hide the input field(not all the component but the input itself) and display the ellipse text on blur and on focus I have to revert back to show the input field so then this could not be done by any mean if I just import the Input Component 

I hope this will help you understand the approach

osamajandali: I would I hack it with css while I can create separate components and reuse the same styles. We just need to remove the code duplicates, but that will be later on when we restructure the input better.

"Sorry i dont know what you mean here, do you have an example of any screen shots? TY"
if you take a look on the design of the input you will find the text is ellipsised 
image
in the first glance you will think easy just make the text ellipsis and put it in the input value. But think of a use case where someone is using it in a form and you submit the form, what value you will get? the ellipsised wallet address not the actual wallet address
So the case here is you only need to display wallet address ellipsis without changing the input value, and the solution to that is to create a text element that holds the ellipsis text and hide the input and when the user want to edit hide that element and show the input back
that would then end up like this 
wallet input

I believe you haven’t checked the code yet so I encourage you to see it and see the recent commits on this PR where I separate the components while having the same styles.

BillyG83: ah nice i had such a busy day, this looks cool, i'll take some item tomorrow to review it again for you

src/components/Input/Input.stories.tsx
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import { ComponentStory, ComponentMeta } from '@storybook/react';
 import Input from '../../components/Input/Input';
+import WalletInputComponent from '../../components/Input/WalletInput';
```
BillyG83, line 4: can you fix both of these imports please I dont know why it went two level up and back down
`../../components/Input/Input` === `./Input` right?

src/components/Input/WalletInput.test.tsx
```diff
@@ -0,0 +1,154 @@
...
+const testValue = '';
+const testPlaceholder = '';
+const testInputId = 'test-input';
+const testLabelId = 'test-label';
+const testDivId = 'test-div';
+const testEllipsisInputId = 'test-ellipsis-input';
+const testBlockieId = 'test-blockie';
+const testEvent = jest.fn();
+
+test("should conditionally render 'empty / filled' className", () => {
```
BillyG83, line 27: niiiiice you used render and screen, i assigned you `<Input />` React18 task next, sorry its beefy but you know Input well now =]

You can close this comment

src/components/Input/WalletInput.tsx
```diff
@@ -0,0 +1,150 @@
...
+    DivWrapperStyled,
+    InputStyled,
+    LabelStyled,
+    StrongStyled,
+    DivStyledEllipsis,
+} from './Input.styles';
+import type { WalletInputProps } from './types';
+import { getEllipsisTxt } from '../../web3utils';
+
+const WalletInput: React.FC<WalletInputProps> = ({
```
BillyG83, line 14: You have done a great job here, but if I am honest I don't love this solution, sorry man I hate to give feedback when its not 100% good, and lets focus on the positive here you have done a great work and learned the kit more. 

The reason I suggested to import `Input` is because now we have two versions of `Input` to maintain. Two prop interfaces, logic and markup. If another dev or two (or more) solved their Input problem like this we would have a mess on our hands. As this is not a critical component for now. Lets park this task and come back to it. If I am honest I'm not very happy with `Input` right now and I might give it some love in the next few days. Then I was thinking to set this as a goal you can work on and we can tick off for your career progress at Moralis. How does that sound? Sorry to be the barer of not good news

BillyG83 (changes requested):
This was a good idea but will make maintenance harder in the long run and may encourage other devs to use the same practise. Your code is all perfect otherwise you have clearly studied and understood the docs which is amazing TY 💚

BillyG83:
thanks for the approval @Mitumaru 👋  😅

BillyG83:
Lets close this one for now @osamajandali and reopen it soon for a 'goal' for you to work towards. For now lets focus on React18 tasks as its causing our community problems

Y0moo:
Looks like it's time for Reopening this PR

PR #404, opened 2022-04-07
React18 refactor banner strip and styles

### LINKED ISSUE #359: React18 refactor BannerStrip
- Remove ReactDOM by using modern RTL setup (see below)
- Scope styles to the component better by not using named exports, but make sure they are not shared in another component as with some niche cases (see below)
- Add this component to the docs, you might get lucky and its already there. There is a bunch of tools to make the gifs, i like Quicktime but a search will prob find even easier tools
- make sure all possible renders / versions of this component are rendered in Storybook, some times we forget and hide great features
- standardise the hell out of the names please, see here for reference https://github.com/web3ui/web3uikit/blob/master/CONTRIBUTE.md#css-naming-an-element

### Test like this
```
import {
    render,
    screen,
    fireEvent,
    act,
    waitFor,
} from '@testing-library/react';

const copyIconTestId = 'this-component';

test('Should render the component', () => {
    render(<ThisComponent />);
    const element = screen.getByTestId(copyIconTestId);
    expect(element).not.toBeNull();

    // Verify that the `icon copy` is visible
    expect(element.textContent).toBe(`${iconTypes.copy} icon`);
});
```

### Scope CSS like this

this is bad, intellisense is going crazy
```
export const ButtonStyled = styled.button`
   // CSS stuff here
`;
```

this is better
```
const ButtonStyled = styled.button`
   // CSS stuff here
`;

export default {
  ButtonStyled,
}
```

then import like this
```
import styles from './ThisComponent.styles';

const { ButtonStyled, } = styles;
```

## need help?
check this PR i already did this for Accordion =]
https://github.com/web3ui/web3uikit/pull/356

AbhinavMV:
- tests meet React18 standards
- all stories are tested
- all layouts / setups are rendered in stories
- scoped CSS without named exports
closes #359 
I have some doubts here.
1. stories file also uses named export. Does that also needs to be scoped?
2. In the section styles, z-index: 1001 is used. But what if we have a navbar and a bannerstrip on the page and user tries to scroll on that page.
In this case the bannerstrip would end up getting on top of navbar. 
Let me know if my understanding is correct or am i missing something. Thanks.

### DIFF
--- a/src/components/BannerStrip/BannerStrip.styles.ts
+++ b/src/components/BannerStrip/BannerStrip.styles.ts
@@ -1,7 +1,23 @@
-import { css } from 'styled-components';
+import styled from 'styled-components';
 import color from '../../styles/colors';
+import fonts from '../../styles/fonts';
+import { BannerStripProps } from './types';
 
-const section = css`
+const getBackgroundColor = (type: string) => {
+    switch (type) {
+        case 'success':
+            return color.green;
+        case 'warning':
+            return color.yellow;
+        case 'error':
+            return color.red;
+        default:
+            return color.blue;
+    }
+};
+
+const SectionStyled = styled.section<Pick<BannerStripProps, 'type' | 'height'>>`
+    ${fonts.text};
     align-items: center;
     color: ${color.white};
     display: flex;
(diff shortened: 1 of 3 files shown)


BillyG83 (approved):
Great job AbhinavMV

1. Scoped stories would have been nice yeah 😅  next time
2. 100% correct this was component no1 and it was fitting needs or the old Admin. This old boy needs some love when the job is done 😁

Issue #406, opened 2022-04-08
Chain Selector - "Select All" button should become "Deselect All" if clicked 


dani69654:
As per design request, if the "Select All" button is clicked, it should become "Deselect All".

![Screenshot 2022-04-08 at 13 59 57](https://user-images.githubusercontent.com/60792849/162431342-9ede54a1-b434-4feb-a305-3010ce91788a.png)

osamajandali:
I will take care of this

BillyG83:
@osamajandali nice work bro TY for taking the initiative here

please remember to use the `in progress` label and make your branch wit the `Development` tab in the sidebar there so the task tracks your work TY. I will add this to the handbook actually

PR #426, opened 2022-04-12
feat: Chain selector deselect

osamajandali:
Refer to issue #406 

Added functionality to toggle into deselect all once all the cards are selected and vice versa

### DIFF
--- a/src/components/ChainSelector/ChainSelector.test.tsx
+++ b/src/components/ChainSelector/ChainSelector.test.tsx
@@ -43,4 +43,7 @@ test('renders all the cards', async () => {
     expect(getByText('Goerli')).toBeDefined();
 });
 
-// TODO: Use userEvent testing after React18 upgrade
+xtest('Batch select/deselect is working', () => {
+    // TODO: test select/deselect toggle
+    // TODO: test select/deselect all cards
+});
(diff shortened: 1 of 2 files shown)


src/components/ChainSelector/ChainSelector.test.tsx
```diff
@@ -42,3 +42,5 @@ test('renders all the cards', async () => {
     expect(getByText('Ropsten')).toBeDefined();
     expect(getByText('Goerli')).toBeDefined();
 });
+
+// TODO: Use userEvent testing after React18 upgrade
```
BillyG83, line 46: I had a thought when i was talking to @AbhinavMV ...

Instead of adding a TODO can you this?

```
xtest('name of test', () => {
  // TODO what to test
});
```

notice `xtest` this means the test runner adds a skipped test, this will bring our attention to it later

BillyG83 (changes requested):
sorry i had a good idea about a skipped test rather than a TODO since we last spoke. please update then i will test this locally in Storybook =]

src/components/ChainSelector/ChainSelector.test.tsx
osamajandali: Didn't know about this xtest
Nice one! thank you

PR #450, opened 2022-04-14
feat: New Button theme: text

niklabh:
Screenshot from 2022-04-14 12-02-51

Screenshot from 2022-04-14 12-03-03

Screenshot from 2022-04-14 12-03-21

### DIFF
--- a/src/components/Button/styles/standardThemes.ts
+++ b/src/components/Button/styles/standardThemes.ts
@@ -160,6 +160,7 @@ export const text = css`
     background-color: transparent;
     border-color: transparent;
     color: ${color.blue};
+    display: inline-flex;
 
     :hover {
         color: ${color.paleCerulean};
--- a/src/components/Button/styles/standardThemes.ts
+++ b/src/components/Button/styles/standardThemes.ts
@@ -157,14 +157,13 @@ export const translucent = css`
 `;
 
 export const text = css`
-    background-color: ${color.white};
+    background-color: transparent;
     border-color: transparent;
-
     color: ${color.blue};
 
     :hover {
         color: ${color.paleCerulean};
-        background-color: ${color.blueLight};
+        background-color: ${color.blueSkyDark};
         border-color: transparent;
 
         svg {
@@ -174,6 +173,7 @@ export const text = css`
 
     :active {
         border-color: transparent;
+        background-color: ${color.grey};
     }
 
     :focus {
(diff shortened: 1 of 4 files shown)


src/components/Button/styles/standardThemes.ts
```diff
@@ -155,3 +155,32 @@ export const translucent = css`
         fill: ${color.white};
     }
 `;
+
+export const text = css`
+    background-color: ${color.white};
```
Y0moo, line 160: Let's replace it to `transparent` here

src/components/Button/styles/standardThemes.ts
```diff
@@ -155,3 +155,32 @@ export const translucent = css`
...
+        color: ${color.paleCerulean};
+        background-color: ${color.blueLight};
+        border-color: transparent;
+
+        svg {
+            fill: ${color.paleCerulean};
+        }
+    }
+
+    :active {
```
Y0moo, line 175: Following the style in Figma, we need to have a "gray" bg in active state

image

Y0moo (changes requested):

src/components/Button/styles/standardThemes.ts
```diff
@@ -155,3 +155,32 @@ export const translucent = css`
...
+
+export const text = css`
+    background-color: ${color.white};
+    border-color: transparent;
+
+    color: ${color.blue};
+
+    :hover {
+        color: ${color.paleCerulean};
+        background-color: ${color.blueLight};
```
Y0moo, line 167: Should be there `background-color: ${color.blueSkyDark};`

Y0moo:
Great work @niklabh, please check my comments

niklabh:
Done

BillyG83 (changes requested):
nit sure this works, it looks like `secondary` when i use `text`

what is the required result? Just text but acts like a button? Maybe it would need less padding also to sit inline with text?

BillyG83:
Screenshot 2022-04-18 at 13 52 12

from my testing

Y0moo:
weird, why do you have themes as a string input and not as a group of radios in the canvas? It should be like: 

image

niklabh:
> weird, why do you have themes as a string input and not as a group of radios in the canvas? It should be like:
> 
> image

Strange. No idea why its happening

BillyG83:
> why do you have themes as a string input and not as a group of radios in the canvas? It should be like:

i think once you have too many options Storybook shows the dropdown instead of radios to save space

PR #456, opened 2022-04-14
Add Color Interface

### LINKED ISSUE #74: color interface
Pretty simple task for Typescript fans, please add a color interface for all components to be passed various color props

- only use colors from `/styles/colors.js`
- make a new `/interface` folder there is more that can be centralised like Input validations and others if you want to do that too, great.

**MORE INFO:**

To be more clear i would expect this end result, sorry for the confusion i caused.

```
<AnyComponent bgColor={green} textColor={white}
```

AnyComponent: don't worry about what component it is and also you do not need to add bgColor or textColor props to every component. Instead imagine that this component is completed and in our kit, plus it has these props setup and ready to use.

When setting bgColor I should use the interface we want built in this task, so it can only be set to the string names of the colors we use in `/styles/colors.js` and it will return the correct HEX. 
So 'green' = '#68738D'

AbhinavMV:
- Added an interface for colors.
- Added an optional type(bgColor) in NewComp to show how to use this component. (This might help others know how to use this type. Let me know if this needs to removed)

image

closes #74

### DIFF
--- a/src/components/NewComp/NewComp.styles.ts
+++ b/src/components/NewComp/NewComp.styles.ts
@@ -10,7 +10,7 @@ import fonts from '../../styles/fonts';
 import type { NewCompProps } from './types';
 
 // pick the props that you will use for conditional CSS
-type TStyleProps = Pick<NewCompProps, 'state' | 'hasUnderline'>;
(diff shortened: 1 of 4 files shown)


BillyG83 (approved):
I cant believe you did this, let me plan around with this and think how best to extend it

PR #467, opened 2022-04-17
feat: Add isRow prop to radio button

### LINKED ISSUE #465: Radio display as row
Screenshot 2022-04-17 at 17 48 51

- to match the design please add an optional boolean prop isRow
- flex the radios but add `flex-wrap: wrap` in case the text is long and the screen is small

AbhinavMV:
Things done.
- Added isRow optional prop. default set to false.
- Added styles for the functionality
- Sort lines in CSS
closes #465 
Small screen
image

Large Screen
image


BillyG83 (approved):
perfect TY @AbhinavMV

PR #471, opened 2022-04-18
feat: fix for Radio Button

rayyan224:


src/components/Radios/Radios.tsx
```diff
@@ -22,13 +22,20 @@ const Radios: React.FC<RadiosProps> = ({
...
 }) => {
-    const formattedID = id.replace(/\s/g, '-');
+    const formattedID = id?.replace(/\s/g, '-');
     const isCreditCards = Boolean(typeof items[0] === 'object');
     const [whichIsChecked, setChecked] = useState<number>(
         setWhichIsChecked || items.length,
     );
 
+    useEffect(() => {
+        if (typeof selectedState != 'undefined') {
```
Y0moo, line 34: Nice solution. This helps to avoid problems with the `0` number

Y0moo:
Why do we need this prop? We have `setWhichIsChecked`, doesn't it cover this functionality?

Y0moo (changes requested):
Why do we need this prop? We have setWhichIsChecked, doesn't it cover this functionality?

src/components/Radios/Radios.tsx
Y0moo: Since we already have `setWhichIsChecked` we need to replace `selectedState` to `setWhichIsChecked` here

rayyan224: Discussed on Slack, this is a breaking a change, merging current solution

Y0moo (approved):

PR #482, opened 2022-04-22
feat: add input prefix for input icons

rayyan224:
Added icon prefix type for inputs
Added ability to change input into, this allows for injecting refs from client side code


dani69654:
Looks good @rayyan224

PR #496, opened 2022-04-26
feat : added reverse resolution

0xPr0f:
Added reverse resolution and mini ticker (needs to be updated) to the connectButton component

### DIFF
--- a/src/components/ConnectButton/ConnectButton.tsx
+++ b/src/components/ConnectButton/ConnectButton.tsx
@@ -24,7 +24,7 @@ const ConnectButton: React.FC<ConnectButtonProps> = ({
     chainId,
     moralisAuth = true,
     signingMessage = 'Moralis Authentication',
-    resolveAddress = true,
+    resolveAddress = false,
 }) => {
     const {
         web3,
--- a/src/components/ConnectButton/ConnectButton.tsx
+++ b/src/components/ConnectButton/ConnectButton.tsx
@@ -112,7 +112,7 @@ const ConnectButton: React.FC<ConnectButtonProps> = ({
         if (isInitialized) logout();
     }
     async function resolve() {
-        if (resolveAddress === true) {
+        if (resolveAddress) {
             const name = await web3?.lookupAddress(account!);
             if (name) {
                 setName(name!);
--- a/src/components/ConnectButton/ConnectButton.tsx
+++ b/src/components/ConnectButton/ConnectButton.tsx
@@ -115,20 +115,13 @@ const ConnectButton: React.FC<ConnectButtonProps> = ({
         if (isInitialized) logout();
     }
     async function resolve() {
-        const network = await web3?.getNetwork();
-        const chainID = network?.chainId;
-        if (chainID === 1 && resolveAddress === true) {
-            try {
-                let name = await web3?.lookupAddress(account!);
-                console.log('passed');
-                if (name !== null || undefined || '') {
-                    setHasName(true);
-                    setName(name!);
-                } else {
-                    setHasName(false);
-                }
-            } catch (e) {
-                console.log('error', e);
+        if (resolveAddress === true) {
+            let name = await web3?.lookupAddress(account!);
+            if (name !== null || undefined || '') {
+                setHasName(true);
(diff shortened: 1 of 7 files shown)


0xPr0f:
image

Currently how the feature looks

0xPr0f:
Added functionality to customise name of the components 
(They have default names when not customised )
ezgif-3-15fbff1bb2

src/components/ConnectButton/ConnectButton.tsx
```diff
@@ -100,16 +115,29 @@ const ConnectButton: React.FC<ConnectButtonProps> = ({
         deactivateWeb3();
         if (isInitialized) logout();
     }
-
+    async function resolve() {
+        if (resolveAddress === true) {
```
locothedev, line 119: can be simplified
`if(resolveaddress)`

src/components/ConnectButton/ConnectButton.tsx
```diff
@@ -100,16 +115,29 @@ const ConnectButton: React.FC<ConnectButtonProps> = ({
...
+            let name = await web3?.lookupAddress(account!);
```
locothedev, line 120: never use let or var if you are not reassigning a value. Use const

src/components/ConnectButton/ConnectButton.tsx
```diff
@@ -100,16 +115,29 @@ const ConnectButton: React.FC<ConnectButtonProps> = ({
...
+            if (name !== null || undefined || '') {
```
locothedev, line 121: can be simplified
`if (name)`

src/components/ConnectButton/ConnectButton.tsx
```diff
@@ -100,16 +115,29 @@ const ConnectButton: React.FC<ConnectButtonProps> = ({
...
+                setHasName(true);
+                setName(name!);
+            } else {
+                setHasName(false);
+            }
+        } else {
+            setHasName(false);
```
locothedev, line 128: instead of else just return after `setName();`

src/components/ConnectButton/types.ts
```diff
@@ -1,4 +1,9 @@
 export interface ConnectButtonProps {
+    /**
+     * name on the button, by default it is 'Connect Wallet'
+     */
+    name?: string;
```
locothedev, line 5: customText would be better here!

src/components/NativeBalance/NativeBalance.tsx
```diff
@@ -2,16 +2,33 @@ import React, { useEffect, useState } from 'react';
...
 
 const NativeBalance: React.FC<NativeBalanceProps> = ({ style }) => {
     const { account, chainId, web3, Moralis } = useMoralis();
     const [balance, setBalance] = useState<{
         formatted?: string;
         balance?: unknown;
+        ticker?: string;
     }>({});
 
+    function ticker(chain: string) {
```
locothedev, line 15: we have `getNativeByChain` in `src/web3uitils/networks.ts`

https://github.com/web3ui/web3uikit/blob/master/src/web3utils/networks.ts#L85

locothedev (changes requested):
thank you

src/components/NativeBalance/NativeBalance.tsx
0xPr0f: ok, did not see this, will edit

src/components/ConnectButton/ConnectButton.tsx
0xPr0f:
```js
 if (name) {
                setName(name!);
                return setHasName(true);
            }
            return setHasName(false);
```
I did this, dont know if it is what you had in mind

i have to revert the state so it will render the address and not the name

BillyG83:
hey i added you here @Y0moo because i knew you would be interested to take a look 
this looks cool @0xPr0f TY🏅

BillyG83:
hey @locothedev if you are happy can you close your review please bud, it blocks the merge

and @Y0moo can i ask for your final approval here then please merge for us when you are happy

PR #500, opened 2022-04-26
input match validation

dani69654:
A validator to check if the input matches a given string

https://user-images.githubusercontent.com/60792849/165331591-68823980-1140-4758-87cf-e2d691b02ca0.mov

### DIFF
--- a/src/components/Input/Input.stories.tsx
+++ b/src/components/Input/Input.stories.tsx
@@ -196,3 +196,25 @@ InputFilledHack.args = {
     name: 'Test text Input',
     placeholder: 'Yeah, you can use placeholder now',
 };
+
+export const CustomInput = FilledTemplate.bind({});
+CustomInput.args = {
+    label: 'Custom input',
+    name: 'Test text Custom input',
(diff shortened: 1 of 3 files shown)


src/components/Input/Input.stories.tsx
```diff
@@ -197,6 +197,27 @@ InputFilledHack.args = {
...
+    // ),
+};
+
+export const ValidateStringsMatch = Template.bind({});
+ValidateStringsMatch.args = {
+    label: 'The input should be `Moralis`',
+    type: 'text',
+    validation: {
+        match: 'Moralis',
+        matchInvalidMessage: 'The input is invalid',
```
BillyG83, line 218: should have made this funnier 👎

AbhinavMV:
@dani69654  
The same functionality can be achieved using below value in validation prop
```
{
  "regExp": "Moralis"
}
```
And giving value in errorMessage prop 
Hope i am not missing anything here :)

BillyG83:
it looks good to me but the tests failed

PR #533, opened 2022-04-29
477 pricing plan card does not much designs

rayyan224:
Screen Shot 2022-04-29 at 6 06 10 PM
Screen Shot 2022-04-29 at 6 05 59 PM


akinolu52:
Nice work, some questions

1. how do you handle the subtile close to the title text?

PR #536, opened 2022-04-30
[feat]: Verify component

AbhinavMV:
verifycode

Things shown in the gif above:
1. Typing the code.
2. Deleting the typed code with backspace.
3. Copy pasting the code.

I have take most things into consideration. Please let me know if i missed any use case. 💪🏻 
Also i have just 1 design related issue with this.
How should i handle mobile design? with 5 code input it looks a bit odd for mobile and if length is increased it would be worse. 😫 
issueVerify


AbhinavMV:
Sorry messed up here will commited on a wrong branch will raise fresh PR soon.

PR #537, opened 2022-04-30
Fix modal v scroll

### LINKED ISSUE #510: [Modal] -  The content scrolls vertical is the modal does not fit in the viewport
@osamajandali The bug discussed on Slack

osamajandali:
Fix #510 
- Fixed the vertical scrolling for big content modal so it scrolls outside the modal not inside

https://user-images.githubusercontent.com/13832392/166104665-5d7b6dd1-d8b5-4f7d-9c87-8892e00831b1.mp4

- I realized that there are some modals that are not vertical centered and others with vertical center like this modal
image

So I added `isCentered` props to center the modal vertically on demand

### DIFF
--- a/src/components/Modal/Modal.stories.tsx
+++ b/src/components/Modal/Modal.stories.tsx
@@ -11,9 +11,17 @@ import Dropdown from '../Dropdown/Dropdown';
 import { Typography } from '../Typography';
 import color from '../../styles/colors';
 
+const hasPositionAbsoluteFix = {
+    transform: 'scale(1)',
+    height: '90vh',
+};
+
 export default {
     title: '5.Popup/Modal',
     component: Modal,
+    decorators: [
+        (storyFn) => <div style={hasPositionAbsoluteFix}>{storyFn()}</div>,
+    ],
 } as ComponentMeta<typeof Modal>;
 
 const Template: ComponentStory<typeof Modal> = (args) => {
--- a/src/components/Modal/Modal.stories.tsx
+++ b/src/components/Modal/Modal.stories.tsx
@@ -11,38 +11,32 @@ import Dropdown from '../Dropdown/Dropdown';
 import { Typography } from '../Typography';
 import color from '../../styles/colors';
 
-const hasPositionAbsoluteFix = {
-    transform: 'scale(1)',
-    height: '90vh',
-};
-
 export default {
     title: '5.Popup/Modal',
     component: Modal,
-    decorators: [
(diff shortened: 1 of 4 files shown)


src/components/Modal/Modal.stories.tsx
```diff
@@ -11,38 +11,32 @@ import Dropdown from '../Dropdown/Dropdown';
 import { Typography } from '../Typography';
 import color from '../../styles/colors';
 
-const hasPositionAbsoluteFix = {
-    transform: 'scale(1)',
-    height: '90vh',
-};
-
 export default {
```
AbhinavMV, line 14: Instead of completely deleting this feature. 
Lets give an optional prop that lets the developer decide if he/she wants this or not.

osamajandali: This was in the storybook for viewing only
It was used only to fix the absolute view inside storybook

AbhinavMV: ohh cool. 
LGTM 🚀

src/components/Modal/Modal.stories.tsx
```diff
@@ -11,38 +11,32 @@ import Dropdown from '../Dropdown/Dropdown';
...
     title: '5.Popup/Modal',
     component: Modal,
-    decorators: [
```
BillyG83, line 22: why do you want to remove this fix? Without it the docs are totally broken, see?

Screenshot 2022-05-02 at 16 54 14

src/components/Modal/Modal.stories.tsx
```diff
@@ -67,6 +61,23 @@ Regular.args = {
     ],
 };
 
+export const VerticalCenter = Template.bind({});
```
BillyG83, line 64: why is `isCentered` a prop? I would assume the Modal should always aim to be the the center of the screen horizontally and vertically by default. Is that not the case? Maybe you have some good evidence here perhaps? TY

BillyG83 (changes requested):

src/components/Modal/Modal.stories.tsx
osamajandali: As I wrote in the PR description, I found in the design so of the modals to be not centred and some are centred 
Plus not centering it will help in the scrolling behaviour

src/components/Modal/Modal.stories.tsx
osamajandali: Maybe I removed it by wrong can't remember why I removed it
Will add it back

dani69654 (approved):

PR #539, opened 2022-04-30
[feat]: Verify code component

### LINKED ISSUE #463: Verify Code component
we need a new component for verifying codes:

Inputs
- sting for label text
- number for amount of inputs to render
- type for number or text versions
- onChange function that returns the values concatenated into a string (empty inputs are ignored)

check list
- [ ] Made from NewComp to ensure standardisation
- [ ] Has initial, focus, filled states like in design please
- [ ] Using modern RTL (see Accordion)
- [ ] [actions](https://storybook.js.org/docs/react/essentials/actions/) returning the value in Storybook
- [ ] killer DX

design
Screenshot 2022-04-17 at 15 53 02

being used in context
Screenshot 2022-04-17 at 15 43 28

AbhinavMV:
verifycode

Things shown in the gif above:
1. Typing the code.
2. Deleting the typed code with backspace.
3. Copy pasting the code.

I have taken most things(onfocus,onblur,autofocus) into consideration. Please let me know if i missed anythings. 💪🏻 
closes #463


AbhinavMV:
verifycode

Refactored the component using input tags
OnCompleted now fires once - after all data is input , and after copy paste
Removed onchange function.

BillyG83 (approved):
amazing work here @AbhinavMV i made some minor updates if you want to check the final commit, but wow you really built something amazing here. Although i removed some bugs, like you could not change the values, i also added another as the max length does not work for number inputs. However i think with your focus feature its much harder to go wrong now

AbhinavMV:
Thanks Bill
I went with the text type instead of number as it was making things easier 😋
Will look at your commit and learn from it. Thanks a lot again for all those suggestions really helped to simplify this component 🙌

PR #540, opened 2022-04-30
477 pricing plan card your plan toggle

rayyan224:
Screen Shot 2022-04-30 at 1 01 33 PM


Y0moo:
image

Y0moo:
We don't need a responsive card in case we will use it as an item for the grid with `auto-fit`

rayyan224:
Now Responsive ... merging
Screen Shot 2022-04-30 at 1 32 24 PM

PR #401, opened 2022-04-07
400 linkto and modal more responsive

BillyG83:
Modal button fix
Screenshot 2022-04-07 at 13 15 40

Modal width fix
Screenshot 2022-04-07 at 13 15 57

LinkTo long text fix
Screenshot 2022-04-07 at 13 22 15

PR #543, opened 2022-05-02
405 hero responsive

SmithVyne:
Closes #401 

I have also added a story that demonstrates how we can add images to Heros better by passing them as children rather than using a prop. So let's leave all additional child elements that a Hero needs to be included as a child component. Devs can style and modify components better and suit to their individual needs.

**Use this:** 
image
**Instead of this:**
image


SmithVyne:
You are always welcome 💪

PR #551, opened 2022-05-03
feat: description for select comp

### LINKED ISSUE #549: [Select] Add description prop
Add optional desciption prop to Select just like we currently have for Input.

AbhinavMV:
Description below select component. for both Traditional HTML5 and default. 
For default if state=error and errorMessage is present error message would be shown else description message.

image

closes #549


SmithVyne (approved):
Great job.

PR #556, opened 2022-05-04
Fix: Set box sizing to border-box to include padding in width.

DanielNgozika:
- Set box-sizing to border-box.


Darrellbor (approved):

BillyG83 (approved):
nice fix dude

PR #561, opened 2022-05-05
259 fix the dropdown select in the button + create dapp

Darrellbor:


src/components/PopoverDropdown/PopoverDropdown.styles.tsx
```diff
@@ -73,15 +74,21 @@ const top = css<TStyleProps>`
...
 `;
 
-const DivStyledArrow = styled.div<TStyleProps>`
-    z-index: 999;
-    ${(p) => (p.position ? getContainerStyleByPosition(p.position) : '')}
+export const DivStyledArrow = styled.div<TStyleProps>`
+    z-index: 1002;
+    width: 130%;
+    height: 10px;
+    top: 32px;
```
akinolu52, line 88: nice work @Darrellbor 

there's this small issue on the popoverdropdown; it has an extra space between the arrow the dorpdown list to fix this issue; do this:

`top: 34px`

Screenshot 2022-05-06 at 13 31 30

akinolu52 (approved):
nice work @Darrellbor , however there's a small fix - check the review

src/components/PopoverDropdown/PopoverDropdown.styles.tsx
BillyG83: nice catch @akinolu52 and very cool to provide a fix too. Good collab guys i live to see this 🤩

BillyG83 (changes requested):
I added a story to test this bug but it seems broken, can you see the tail of the Popover in this pic? There is actually two of them. Can you you take another look please? I think the story i added here will make it easier to debug

Screenshot 2022-05-10 at 13 49 51

Darrellbor:
Alright will do.
how do i replicate it though?

osamajandali:
@Darrellbor I see Bill added that story so no need to replicate it.

I encourage you to go through the whole css as I see there are many things need to be fixed but mainly values changes to the arrow will do it all I think.

As you can see there are two arrows here, that's because we have arrow components and another on in `:after`..so I think it's better to remove the component and keep everything clean in `:after`..

Darrellbor:
@osamajandali i don't think i really get this part.

osamajandali:
Okay let me explain it more

There are two problems in here:
- **The position of the arrow goes wrong sometimes**: solution to that is test all positions and adjust the values properly 
- **There are two arrows (as you saw in bill's screenshot)**
if you look at [this line ](https://github.com/web3ui/web3uikit/blob/66bceba114ecd783d8676c4b30a7e35d5bb12ca7/src/components/PopoverDropdown/PopoverDropdown.tsx#L47) you will find that there is a component called `DivStyledArrow`
and if you check [this line](https://github.com/web3ui/web3uikit/blob/66bceba114ecd783d8676c4b30a7e35d5bb12ca7/src/components/PopoverDropdown/PopoverDropdown.styles.tsx#L27) you will find this is some code written to make an arrow with css and it's been written in `:after` pseudo element 
So you need to go for either on of them and remove the other. I suggest to for the one in `:after` pseudo element as then it will move with the parent element and it will make it less buggy.

Let me know if I made it clear here

AbhinavMV:
Hey @Darrellbor don't know if it's already mentioned. But just noticed one small bug.

image
In all the stories. the last popover element has an arrow which is visible on hover. (Just like in the above image)

Darrellbor:
@osamajandali @AbhinavMV I spent a better part of yesterday trying to repurpose this component but i believe at this point it will be best if this component is rebuilt in version 2, i spoke with @BillyG83 about this too.

Darrellbor:
Sure thing man @osamajandali 
Totally.

osamajandali:
Will close this for now need a refactor as you said

PR #563, opened 2022-05-06
449: onTab feature added to the different button themes

Darrellbor:


BillyG83 (changes requested):
this is all good, nice work. I didn't have time to double check the designs, i trust you were careful. But there is a few you missed

Screenshot 2022-05-06 at 16 06 02

if you can update these also and check the values are correct as I am sure you did for the others, that would be perfect TY. Touchdown scored, now kick the extra points =]

Darrellbor:
@BillyG83 
I have updated the colored buttons onTab state.
**Note:** Button themes of type **link**, **text**, and **ghost** currently have no onTab state.

PR #582, opened 2022-05-11
[577]:... typography and font responsiveness

Darrellbor:
- New breakpoint file created
- New fontBreakpoint file created to cater for the different media queries for heading tags
- Responsiveness is slated for headers h1 through h6

### DIFF
--- a/src/components/Typography/Typography.styles.ts
+++ b/src/components/Typography/Typography.styles.ts
@@ -1,5 +1,6 @@
 import { css } from 'styled-components';
 import color from '../../styles/colors';
+import fontBreakpoint from '../../styles/fontBreakpoint';
 import type { TypographyProps, variantType, weightType } from './types';
 
 const openSans = css`
@@ -57,6 +58,7 @@ const h1 = css`
     font-size: 36px;
     line-height: 40px;
     letter-spacing: -0.75px;
+    ${fontBreakpoint.h1};
 `;
 
 const h2 = css`
@@ -64,6 +66,7 @@ const h2 = css`
     font-size: 32px;
     font-weight: 400;
     line-height: 36px;
+    ${fontBreakpoint.h2}
 `;
 
 const h3 = css`
@@ -71,6 +74,7 @@ const h3 = css`
     font-size: 24px;
     font-weight: 400;
     line-height: 32px;
+    ${fontBreakpoint.h3}
 `;
 
 const h4 = css`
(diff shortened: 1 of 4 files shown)


src/styles/breakpoints.ts
```diff
@@ -0,0 +1,26 @@
...
+const breakpoints = {
+    xxl: '1400px',
+    xl: '1200px',
+    lg: '1025px',
+    md: '768px',
+    sm: '576px',
+    xs: '320px',
+};
+
+export const fontSize = {
```
BillyG83, line 18: i like this approach but i think i would prefer this object to be in `styles/fonts.js`
what do you think?

BillyG83 (approved):
i think this is great @Darrellbor might you consider my feedback please? 🙏 
DM me on Slack when you want this merged 👍

src/styles/breakpoints.ts
Darrellbor: I believe that also works nicely, 
somewhere in my mind, I thought because this would be used across board on the application, it might want to be a file of its own.

Darrellbor: I reverted to having them in the breakpoint.ts file to avoid a cyclical dependency issue between font.ts and fontBreakpoint.ts

osamajandali: @Darrellbor can you please explain to me what is cyclical dependency issue?

osamajandali (commented):
Would like to know why we are having `fontBreakpoint` and `fonts.js` why aren't they combined together ?

Darrellbor:
@osamajandali there was no particular reason. I just thought it to be a good seperation of concern.

src/styles/breakpoints.ts
Darrellbor: Exporting from `fonts.ts` and importing into `fontsBreakpoint.ts` and vise versa

osamajandali:
@Darrellbor From what I understand there is no separation because you re-include it again in `font.ts` 
It's good idea to keep it simple in `font.ts`

Darrellbor:
Alright noted

Darrellbor:
@osamajandali adjusted appropriately.

src/components/Typography/Typography.styles.ts
```diff
@@ -1,5 +1,6 @@
 import { css } from 'styled-components';
 import color from '../../styles/colors';
+import fontBreakpoint from '../../styles/fontBreakpoint';
```
AbhinavMV, line 3: Please update this import as well

Darrellbor: Oh, i already noticed the issue and pushed a fix for it 
Thanks

AbhinavMV (approved):
Great work 🙌🏻 
LGTM 🚀

PR #584, opened 2022-05-12
[583]:... input on valid change prop

Darrellbor:
### On Valid Change Implementation
- Seclude the validation
- handle onBlur Separately
- handle onValidChange from valueChange
1.  if Valid return input 
2. else if not valid return undefined
- Implemented this across checkbox, radios, textarea, and select.


Darrellbor:
@BillyG83 I just updated the onValidChange to not fire if the input into the selectors is not valid.

AbhinavMV:
Hey @Darrellbor 
Awesome work here 🙌🏻 

Just have 1 question
> Input returns normally onChange (no breaking changes)
> Input returns data with onValidChange if the change passes the regex of the Input

According to issue description onChange should be returned always and when regex passes then onValidChange should also be returned right? Right now only onValidChange is being returned in actions 

(Please also resolve the conflicts 😄)

Darrellbor:
@AbhinavMV 
So basically for this issue 
1. onChange; value is always returned regardless of the regex or validation rules 
2. onValidChange is only called on valid validation rules that include the regex rules.

AbhinavMV:
Thanks for explaining @Darrellbor 
I had that doubt because storybook action was not showing onChange action.  

Also i found another bug i think.
input

Current -> This gif shows the input component with state prop = error, after clicking on it and stepping off(onBlur) from it, without doing anything else. The state changes to confirm.

Expected -> if state prop is given error value, it should remain in error until we change it.

PR #588, opened 2022-05-13
Commented out table logging

pgrignaffini:
While in console mode on dev tools a table keeps get logged making difficult to find the console.log information from progra
Screen Shot 2022-05-12 at 10 44 41
m


src/components/Button/Button.styles.ts
```diff
@@ -100,7 +100,7 @@ const getSizeStyles = (size: string) => {
 };
 
 const getColored = (color: string, theme: string) => {
-    console.table(coloredRed);
+    //console.table(coloredRed);
```
BillyG83, line 103: be brave, delete it please

PR #590, opened 2022-05-15
554 notification success and error notification icons should be coloured

DanielNgozika:
Make all the icon colours match their respective notification type.


Y0moo:
GJ on this PR @DanielNgozika

PR #595, opened 2022-05-18
fix: invalid useState params were setting an empty array

dani69654:


dani69654:
```ts
// OLD
const o = [...Array(length)].map(() => '')
console.log('OLD::', o)
// NEW
const n = new Array(5).fill('')
console.log('NEW::', n)
```
Screenshot 2022-05-18 at 14 13 19

PR #599, opened 2022-05-18
598 verifycode should not allow two digits in one input

dani69654:
https://user-images.githubusercontent.com/60792849/169143993-517c32fe-5b23-4b1c-a5d7-54784e87826c.mov


rayyan224 (approved):
Niceeee 🇮🇹

PR #616, opened 2022-05-24
made todo item unique

akinolu52:
---
name: Unique Todo Items
about: This PR prevents double entries on todo IP items
---

### Checklist

### Solution Description

add all input items into a set and then convert it into an array

### DIFF
--- a/src/components/Todo/Todo.tsx
+++ b/src/components/Todo/Todo.tsx
@@ -40,7 +40,7 @@ const Todo: React.FC<TodoProps> = ({
 
     const addTodo = (e: any) => {
         e.preventDefault();
-        setLists((prevTodo) => [...prevTodo, inputValue]);
+        setLists((prevTodo) => Array.from(new Set([...prevTodo, inputValue])));
         setInputValue('');
         const input: HTMLInputElement | null = document.querySelector(
             `input[key="${key}"]`,


src/components/Todo/Todo.tsx
```diff
@@ -40,7 +40,7 @@ const Todo: React.FC<TodoProps> = ({
 
     const addTodo = (e: any) => {
         e.preventDefault();
-        setLists((prevTodo) => [...prevTodo, inputValue]);
+        setLists((prevTodo) => Array.from(new Set([...prevTodo, inputValue])));
```
rayyan224, line 43: Goood use of sets love it

rayyan224 (approved):

src/components/Todo/Todo.tsx
akinolu52: thanks

rayyan224:
I  think we should do an error to tell the user no duplicates? Or maybe disable button ?

BillyG83 (approved):
smart dude! when @rayyan224 is impressed i know its good CS 😆

src/components/Todo/Todo.tsx
osamajandali: It's a great solution indeed

Issue #610, opened 2022-05-23
[Table] Add sort column


dani69654:
Implement sorting functionality

![image](https://user-images.githubusercontent.com/60792849/169801612-08b00e22-1def-4112-9df2-cfcf16b7a75a.png)

PR #619, opened 2022-05-26
feat: add sorting columns for table

AbhinavMV:
---
name: 'Table Component'
about: Sorting columns
---

### Checklist

### Issue Description

Related issue: #610 

### Solution Description

Added a prop that would take boolean array input. True if column is sortable else false.
On clicking on appropriate table header the table would be sorted.
Please checkout Table-> Sync Table example story

### DIFF
--- a/src/components/Table/Helper.ts
+++ b/src/components/Table/Helper.ts
@@ -5,7 +5,7 @@ export function paginate(
     maxPages: number = 10,
 ): number[] {
     // calculate total pages
-    let totalPages = Math.ceil(totalItems / pageSize);
+    const totalPages = Math.ceil(totalItems / pageSize);
 
     // ensure current page isn't out of range
     if (currentPage < 1) {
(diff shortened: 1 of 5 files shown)


osamajandali:
I liked the approach so much! Great job @AbhinavMV 

My only concern is that this client side sorting only and this approach won’t work for server side sorting, have you thought about this or you indented to make it client side sorting only?

let me know if you need more clarification of what I said.

AbhinavMV:
Thanks @osamajandali 

Hmm i was only going for client side sorting. 
Usually for tables sorting is done when user clicks on headers, have not seen many usecases for server side sorting here.

Let me know if you think it would be a helpful feature, will be happy to learn and implement it 💪🏻

osamajandali:
Thanks for explaining @AbhinavMV 
I think we can go with this one at the moment and add the server side later.

Just FYI why I think about server side because of those scenarios: 

- Imagine a table of 1000 or maybe 10,000 columns which is a lot of data, so then for sure the request won't get you all of those data at one request, there should be a way for lazy loading or a pagination which on every page of the table you request the next 10/20 records..so then you don't have all the 10k data where you can perform the sorting on them.
- The second scenario is that, during the integration the api maybe have a pagination on it which lead you then to call the request multiple times to get all the data and then sort them on the client side end.

But again for the most cases client sorting works just fine.

Maybe @BillyG83 can give us more input on this matter

AbhinavMV:
Ohh got it 🙌🏻
Thanks for explaining 
Yes we can add server side later on. 🤝

osamajandali (approved):
Let's gooo 🚀

PR #627, opened 2022-05-30
[Input] Add blockie to input component

AbhinavMV:
---
name: 'Input Component'
about: Add blockie to input component
---

### Checklist

### Issue Description

Adds blockie to input component.

Related issue: #271 

### Solution Description

- Added two new values `address` and `bluredAddress` to type prop.
- When this type is selected Blockie would be visible.
- Also value will compared with address regular expression.
- When blurred address is selected onBlur address would shown as ellipsis text.

Added two stories for this please check those.


osamajandali:
Great job here @AbhinavMV !
I will just need to close this for now as we need to refactor the input as it is becoming more complicated and surely we will reference this code in our refactor.
Thank you!

Issue #664, opened 2022-06-22
Tab Component Design issue


philipehsing:
Tab Component states is not equal to design

**Design Screens**

Default: 
![image](https://user-images.githubusercontent.com/4109592/175030650-fe3a0b30-4a89-4e4f-b8fb-baecaad39273.png)

Hover:
![image](https://user-images.githubusercontent.com/4109592/175030774-c4090824-7401-4f15-ad38-8539ecddda14.png)

Active:
![image](https://user-images.githubusercontent.com/4109592/175030689-789ca387-0497-4baa-855d-c89333b1f702.png)

PR #666, opened 2022-06-22
style: update tab style according to design

philipehsing:
---
name: 'TabComponent'
about: Restyle TabComponent according to design
---

### Checklist

### Issue Description

Related issue: #664

Updated `<Tab />` States styling:
- Default
- Active
- Hover
- Disabled

### DIFF
--- a/src/components/Tabs/Tabs.styles.ts
+++ b/src/components/Tabs/Tabs.styles.ts
@@ -1,6 +1,6 @@
 import styled, { css } from 'styled-components';
 import resetCSS from '../../styles/reset';
-import color from '../../styles/colors';
+import color, { gradientColors } from '../../styles/colors';
 import fonts from '../../styles/fonts';
 
 // Styling Interfaces
@@ -38,7 +38,7 @@ export const StyledTabBar = styled.div<IStyledTabBar>`
     ${(props) =>
         props.haveBackground &&
         `
- background-color: ${color.blueLight};
+ background-color: ${color.culturedBlue};
     border-radius: 16px;`}
 `;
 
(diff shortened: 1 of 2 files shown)


src/styles/colors.ts
```diff
@@ -12,6 +12,7 @@ const color = {
     blueSky: '#9ECCEA',
     blueSkyDark: '#51A5DA',
     blueSkyLight: '#E8EAED',
+    culturedBlue: '#EBEFF9',
```
BillyG83, line 15: could this be better DX (developer experience if it was BlueCultured? The reason is when i type `color.bl` i am hoping to see all the blues listed by intellisence in my code editor. This shade of blue will be hidden unless someone opens the file and checks it manually

src/components/Tabs/Tabs.styles.ts
```diff
@@ -72,23 +72,28 @@ export const BulbTab = styled.div<IStyledBulb>`
...
+            border-color: ${color.blueSky};
+            background: transparent;
         `};
-    ${(props) =>
-        (props.isActive || props.isDisabled) && 'pointer-events: none;'};
+    ${(props) => props.isDisabled && 'pointer-events: none;'};
     ${(props) => props.hasMargin && 'margin-right:6px;'}
-    ${(props) => props.isDisabled && `color:${color.greyDisabled};`}
+    ${(props) =>
+        props.isDisabled && 'opacity: 0.5; background-color: transparent;'}
```
BillyG83, line 88: i think transparent will fix one problem for us, but cause many more for others (maybe even in our admin as tabs are used on white and also on BlueSky i believe

BillyG83 (changes requested):
look good @philipehsing but i have a few comments. We need to always try and think how will any change effect all the developers, not just our Admin software. Nice first PR, let me know what you think about the comments and sorry for being so strict

src/components/Tabs/Tabs.styles.ts
philipehsing: yup, you're right! Let me think about this quick and come up with a solution :)

philipehsing:
@BillyG83 please have a look again :)

BillyG83 (approved):
nice work @philipehsing

Issue #648, opened 2022-06-14
Rebuild Dropdown with no breaking changes


BillyG83:
users are having issues with this component and i think i can remove a lot of JS and replace with CSS

PR #677, opened 2022-06-27
648 rebuild dropdown

BillyG83:
---
name: 'Popover Dropdown with CSS and less JS'
about: no breaking changes but now CSS powered, also depreciation warning function added
---

### Checklist

### Issue Description

Related issue: https://github.com/web3ui/web3uikit/issues/648

### Solution Description

- stripped JS and used more performant CSS
- remove move and moveBody props as we do not need them now
- added depreciation warning util

### DIFF
--- a/src/components/PopoverDropdown/PopoverDropdown.stories.tsx
+++ b/src/components/PopoverDropdown/PopoverDropdown.stories.tsx
@@ -1,170 +1,125 @@
-import { ComponentStory, ComponentMeta } from '@storybook/react';
 import React from 'react';
-import color from '../../styles/colors';
-import colors from '../../styles/colors';
 import PopoverDropdown from './PopoverDropdown';
 import PopoverElement from '../PopoverElement/PopoverElement';
-import { Icon } from '../Icon';
+import color from '../../styles/colors';
+import { ComponentStory, ComponentMeta } from '@storybook/react';
 import { iconTypes } from '../Icon/collection';
 
 export default {
     title: '5.Popup/Popover Dropdown',
     component: PopoverDropdown,
-    subcomponents: { PopoverElement },
-    argTypes: { onClick: { action: 'clicked' } },
 } as ComponentMeta<typeof PopoverDropdown>;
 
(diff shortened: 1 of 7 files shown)


AbhinavMV (commented):
This is awesome code looks so clean now That background color feature is sweet. Love it <3
Depreciated warning is nice and would help in our monorepo.
Just have some thoughts - 
1. Styles on hover are missing (background highlight) or were they removed from design? 
2. removing custom dropdown(move and moveBody) position might break some components in admin panel syncs 
- just a thought instead of custom position we can add more positions like - ‘right-bottom’, ‘left-bottom’ …… Will make it much more controllable.
3. Not sure about this but just FYI
 On smaller screens, this is happening. But since hover state is not there on mobile this might not be problem. What do you think?
image

4. We should style our stories since they would be seen by all users. (This is my OCD talking :P)

BillyG83:
wow wow wow! this is very constructive feedback @AbhinavMV TY so much!!

1. style on hover come from the elements that are passed to it (see PopoverDropdown Element). I am in two minds, should we ask the dev to pass urls and text and we make and style the links... or do we let them pass anything as this component is only concerned with being a dropdown of elements. I am very open to refining and iterating this with your suggestions.

2. It wont 'break' just throw a warning that we are depreciating this feature. I dont think we need it now, in fact i was having trouble working out the value of it, maybe you know better? I thought as i had it always centered no it would not be needed at all. TBH i found them not a good DX, maybe i am wrong on this?

3. Yes this is a problem for this and the previous version, either we offer breakpoints in the component (i do not love this), or devs could be told to detect screen size and set position prop according to the size of the screen. Hover does not work on mobile, but clicking the parent element will show it also

4. This is a big question... do we style it and show CSS the next dev will not need or want, or do we show it in the top left corner where it looks broken. Event with a nice story decorator we have the same problem. Maybe we can research storybook for the best practise?

Checkout BannerStrip story that needs this decorator or it looks broken in Stroybook. Lets see what we can do to improve this

banner story decorator

banner story rendered

So i will merge this (not release until we finish our discussion on point 2) and take it all to our Slack room

Issue #695, opened 2022-07-01
feature: Working on BlockNumber Component


3scava1i3r:
Overview:
I went to various dapps like a uniswap and always saw a blocknumber component in the footer or something so i thought why not add this here.Added an image for reference

![gvg](https://user-images.githubusercontent.com/52003051/176967759-99499ce9-88a2-422e-95b3-1da2199d526e.png)

What I would build:
i would make a small BlockNumber component currently just functional to show the current blocknumber of any evm compatible chain using ethers.

AbhinavMV:
Closed By #696

PR #696, opened 2022-07-01
Feat:blocknumber addedcomp

3scava1i3r:
---
name: 'feat:Added BlockNumber'
about: Pr for issue #695 
---

### Checklist

### Issue Description

Related issue: #695 

### Solution Description

It takes no props just a provider should be available for it to from moralis
final screenshot
dd


AbhinavMV:
This PR also has ENS Avatar changes 
Could you please remove those 🙌🏻

3scava1i3r:
@AbhinavMV  removed the ENSAvatar changes and added prettier to configure my code 
thank you for the help🥂

AbhinavMV:
Looks nice
Could you remove unnecessary console logs 🙌🏻 
How about we use Typography comp to display block number, Devs can then select which type of html tag to use h1,h2.... and also give more styling options.
What do you think? 🤔

3scava1i3r:
surely i would remove all the unnecessary console.logs, I don't have any idea on typography comp.Would you mind giving a link or something for me to look at

AbhinavMV:
@BillyG83 what do you think about returning typography comp instead of span html tag?

https://web3ui.github.io/web3uikit/?path=/story/4-ui-typography--heading-1
@3scava1i3r  You can check here about typography component,

3scava1i3r:
I understand this now,we can add the same in native balance as well

BillyG83:
@AbhinavMV please review and merge if you are happy

AbhinavMV:
Great job here @3scava1i3r 

This looks good. 🚀 
Could you remove console logs and unused code? 
And if you can please also add Typography component (I can add this later if you are having any trouble 🙌🏻 )

3scava1i3r:
I can start the work on typography also will add the same for the native balance one what do you think @AbhinavMV

AbhinavMV:
Sure go ahead @3scava1i3r 
Please note that native balance is used in connect button comp, so please do check there if it breaks there.
If it does leave it for now 🙌🏻

AbhinavMV:
TY @3scava1i3r, Nice job 🔥 🚀

Issue #707, opened 2022-07-06
feature : adding support for ipfs/normal file uploading component


3scava1i3r:
Overview:
Its just a small component to upload any file to moralis or ipfs using moralis.I personally use ipfs for a lot of my nft projects so making a component like this would be great

 
What i'll do:
- Make an input component for addition of files
- Use moralis hooks to add upload support for basic moralis and ipfs

AbhinavMV:
closed by 708

PR #708, opened 2022-07-06
Feat: ipfsinput addcomp

3scava1i3r:
---
name: 'Feat: ipfsinput addcomp'
about: PR to finish  issue #707 
---

### Checklist

### Issue Description

Related issue: #707 

### Solution Description

a screenshot for final product right now,still working on an animation while uploading happens

ss

### DIFF
--- a/src/components/IpfsInput/IpfsInput.stories.tsx
+++ b/src/components/IpfsInput/IpfsInput.stories.tsx
@@ -1,0 +1,24 @@
+import React from 'react';
+import { ComponentStory, ComponentMeta } from '@storybook/react';
+import IpfsInput from './IpfsInput';
+import { moralisContext } from '../../decorators';
+import { IpfsInputprops } from './types';
+
+export default {
+    title: '1.Web3/IpfsInput',
+    component: IpfsInput,
+    decorators: [moralisContext],
+    parameters: {
+        docs: {
+            source: {
+                excludeDecorators: true,
+            },
+        },
+    },
+} as ComponentMeta<typeof IpfsInput>;
+
+const Template: ComponentStory<typeof IpfsInput> = (args: IpfsInputprops) => (
+    <IpfsInput {...args} />
+);
+
+export const Default = Template.bind({});
--- a/src/components/IpfsInput/IpfsInput.tsx
+++ b/src/components/IpfsInput/IpfsInput.tsx
@@ -27,8 +27,8 @@ const IpfsInput: React.FC<IpfsInputprops> = (props) => {
         <>
             <InputButton type="file" name="ipfs" onChange={(e) => {
                 console.log(e)
-                saveFile("e.target.value",
-                    e.target.files[0], { saveIPFS: props.SaveToIpfs })
+                saveFile(String(props.FileName),
+                    e.target.files[0], { saveIPFS: props.SaveToIpfs,type: props.Type })
             }} />
         </>
     );
--- a/src/components/IpfsInput/IpfsInput.tsx
+++ b/src/components/IpfsInput/IpfsInput.tsx
@@ -1,0 +1,37 @@
+import React,{useEffect} from 'react';
+import { useMoralisFile } from 'react-moralis';
+import { IpfsInputStyles } from './IpfsInput.styles';
+import { IpfsInputprops } from './types';
+
+
+/**
+ * An Input buton to upload files to either IPFS or moralis
+ * @param {*} props
+ * @returns <IpfsInput> JSX Element
+ */
+
+const { InputButton } = IpfsInputStyles;
+
+const IpfsInput: React.FC<IpfsInputprops> = (props) => {
+    const { moralisFile,saveFile,} = useMoralisFile();
+
+    useEffect(() => {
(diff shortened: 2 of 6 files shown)


3scava1i3r:
This is still a work in progress😄

3scava1i3r:
new ss
dfjvbd

AbhinavMV:
Issue : No way of knowing the ipfs link
Possible solution: This component also needs an `onComplete: (value)=>void` prop. Here value would be the ipfs link which is returned by the saveFile function. 
This is how i want to use the component.
```jsx
const [ipfsLink,setIpfsLink]=useState("");
<IpfsInput onComplete={(val)=>setIpfsLink(val)} />
```
What do you think ?? If you have any better idea please let me know 🙌🏻

src/components/IpfsInput/IpfsInput.tsx
```diff
@@ -0,0 +1,32 @@
...
+    }, [moralisFile]);
+
+    return (
+        <>
+            <Upload
+                theme={Theme}
+                onChange={(e) => {
+                    console.log(e);
+                    saveFile(String(props.FileName), e, {
+                        saveIPFS: props.SaveToIpfs,
```
AbhinavMV, line 23: Github action failing because `saveFile` second argument type is different.
Used type is - Blob | null
Maybe typecast e to File type

3scava1i3r:
> Issue : No way of knowing the ipfs link Possible solution: This component also needs an `onComplete: (value)=>void` prop. Here value would be the ipfs link which is returned by the saveFile function. This is how i want to use the component.
> 
> ```js
> const [ipfsLink,setIpfsLink]=useState("");
> <IpfsInput onComplete={(val)=>setIpfsLink(val)} />
> ```
> 
> What do you think ?? If you have any better idea please let me know 🙌🏻

Well we do get the response of the same using moralisFile object that i console log.Do you have any suggestions to return this variable?

src/components/IpfsInput/IpfsInput.stories.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+const Template: ComponentStory<typeof IpfsInput> = (args: IpfsInputprops) => (
+    <IpfsInput {...args} />
+);
+
+export const TextOnly = Template.bind({});
+TextOnly.args = {
+    Theme: 'textOnly',
+    OnFinish(value) {
+        console.log(value);
```
AbhinavMV, line 28: Please pass function arguments as follows (vscode shows error right now) - 
```jsx
onFinish: (val)=>{console.log(val)}
```

src/components/IpfsInput/IpfsInput.tsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import { useMoralisFile } from 'react-moralis';
+import { IpfsInputprops } from './types';
+import { Upload } from '../Upload';
+
+const IpfsInput: React.FC<IpfsInputprops> = ({
+    Theme = 'textOnly',
+
+    ...props
```
AbhinavMV, line 9: could you destructure props in type file and give them default values if they are optional 😅

src/components/IpfsInput/types.ts
```diff
@@ -0,0 +1,10 @@
+export interface IpfsInputprops {
+    /**
+     * Set theme  either 'textOnly' or 'withIcon'
+     */
+    Theme?: 'textOnly' | 'withIcon';
+    SaveToIpfs?: boolean;
+    FileName?: string;
+    Type?: string;
+    OnFinish: (value: any) => void;
+}
```
AbhinavMV, line 10: 1. Could you convert these names to camelCase ? SaveToIpfs => saveToIPFS
2. Please also add comments for each type since they are used by storybook.
3. i think `onComplete` is better naming than `onFinish`

src/components/IpfsInput/IpfsInput.tsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import { useMoralisFile } from 'react-moralis';
+import { IpfsInputprops } from './types';
+import { Upload } from '../Upload';
+
+const IpfsInput: React.FC<IpfsInputprops> = ({
```
AbhinavMV, line 6: Could you change this component name to `IPFSInput`.
Sorry for being this picky 😅

src/components/IpfsInput/IpfsInput.tsx
```diff
@@ -0,0 +1,35 @@
...
+    Theme = 'textOnly',
+
+    ...props
+}) => {
+    const { saveFile } = useMoralisFile();
+
+    // useEffect(() => {
+    //     console.log(moralisFile);
+    // }, [moralisFile]);
+
```
AbhinavMV, line 16: Please remove this comment

AbhinavMV (changes requested):

src/components/IpfsInput/types.ts
3scava1i3r: There is a event handler in SaveFile as OnComplete therefore added OnFinish

AbhinavMV (approved):
Awesome job @3scava1i3r 🔥

Issue #750, opened 2022-08-01
[Hero] Refactor and match designs


BillyG83:
Hero has some problems

- hero should take children components like typography and button and lay them out nicely
- hero should take an optional component for image
- Illustration needs updated with this sweet Wizard to be added
- if the image is added please use the layout seen below
- do not use position absolute, use flex and give the image a max width and max height
- it should scale as seen in this design perfectly no matter what content, children or image size is passed

<img width="675" alt="CleanShot 2022-08-01 at 14 24 42@2x" src="https://user-images.githubusercontent.com/13779759/182159592-061c49e8-02b9-4a38-be5f-f46e03cab797.png">

BillyG83:
make it killer DX @AbhinavMV and idiot proof 😁

locothedev:
Would be great if title is optional too

BillyG83:
@locothedev i think it would be best of you passed it children and an optional image component, then you can use whatever text you like

BillyG83:
@AbhinavMV maybe make the title optional for now and we can remove it over time so there is no big braking changes

AbhinavMV:
This is closed by #762

PR #762, opened 2022-08-07
Hero refactor

AbhinavMV:
---
name: 'Pull request'
about: Hero refactor
---

### Checklist

### Issue Description

Related issue: #750 

### Solution Description

image

image


BillyG83 (approved):
hey @AbhinavMV this looks great. we can merge this an open a new ticket to iterate. But i cannot seem to get the align prop to work now. 
Screenshot 2022-08-09 at 09 31 42

Also please add

```
RightContainerDiv:empty { display: none; }
```

then when there is no content it will not show

AbhinavMV:
Nice catch @BillyG83 will add `display: none` styling for no image case 🙌🏻 

align is only for the image now
Since title and subtitle are optional and we might remove them in future iterations.... and alignment of children component should be done by user with div
What do you think ? 🤔

Issue #715, opened 2022-07-10
feature : adding support for slider component in the ui kit


3scava1i3r:
Objective:
Adding a slider component that we usually see 

What I'll do:
I am trying to add a slider component,it is not web3 but seems interesting to add

Reference image below
![image](https://user-images.githubusercontent.com/52003051/178142881-f8a2c41d-abd7-4f6b-9659-ec411f8d5c5e.png)

3scava1i3r:
@AbhinavMV is this still available?PR work almost done

AbhinavMV:
@3scava1i3r You can create a PR for this. TY 🙌🏻

AbhinavMV:
closed as completed #839

PR #793, opened 2022-09-04
Feat slider added component

3scava1i3r:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description

Related issue: #715
### Solution Description

Adding a screenshot of final work

Another small addition was done for IPFSInput where the storybook did not have savetoIPFS boolean set

Screenshot from 2022-09-04 13-41-11


.pnpm-debug.log
```diff
@@ -0,0 +1,23 @@
+{
```
AbhinavMV, line 1: probably delete this file 😅

packages/core/src/lib/Slider/Slider.styles.ts
```diff
@@ -0,0 +1,57 @@
+import styled, { css } from 'styled-components';
+import {color, fonts, resetCSS} from '@web3uikit/styles';
+
+// //////////////////
+// Styled Components
+// CSS Props should be sorted alphabetically
+// //////////////////
+
```
AbhinavMV, line 8: Please remove all the comments

.pnpm-debug.log
3scava1i3r: Ok

packages/core/src/lib/Slider/Slider.styles.ts
AbhinavMV: Please install prettier to remove all the extra whitespace please 😄

packages/core/src/lib/Slider/Slider.styles.ts
```diff
@@ -0,0 +1,57 @@
...
+
+const sliderDiv = styled.div`
+
```
AbhinavMV, line 11: It would be nice if you could change the naming convention.. and follow same as rest of the components
for eg - DivStyled, LabelStyled, InputStyled.....

packages/core/src/lib/Slider/Slider.styles.ts
```diff
@@ -0,0 +1,57 @@
...
+// CSS ORDERING
+// 1 - Imported styles (font, colors, etc)
+// 2 - Normal styles (margin, padding, etc)
+// 3 - Child elements (span, svg, etc)
+// 4 - Pseudo elements (before & after)
+// 5 - Modifier styles (hover, active etc)
+// 6 - State changed styles ${(p) => (p.prop ? green : red)};
+// each should be sorted alphabetically as best as possible
+
```
AbhinavMV, line 51: Remove all the unnecessary comments please

packages/core/src/lib/Slider/Slider.test.tsx
```diff
@@ -0,0 +1,202 @@
...
+import {color,rgbToHex} from '@web3uikit/styles';
+
+// importing testID from button and icon
+import { buttonTestId } from '../Button/Button.test';
+
+const iconTestId = 'test-icon';
+
+// importing my stories to test
+const { Default, InitializeRed, UnderLinedText } = composeStories(stories);
+
```
AbhinavMV, line 20: Seems like this test file is not updated 
Could you add some valid basic tests for the component...

packages/core/src/lib/Slider/Slider.tsx
```diff
@@ -0,0 +1,41 @@
...
+    const [currentValue, setCurrentValue] = useState(rangeValue);
+
+    const rangeChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+        const newRangeValue = Number(event.currentTarget.value);
+        setCurrentValue(newRangeValue);
+    };
+
+    return (
+        <styles.sliderDiv>
+            <styles.sliderLabel htmlFor={rangeId}>
```
AbhinavMV, line 23: `sliderDiv` is already de-structured above so you can use it directly instead of `styles.sliderDiv`
Please rename these as mentioned in my above comment 🙌🏻

packages/core/src/lib/Slider/Slider.tsx
```diff
@@ -0,0 +1,41 @@
...
+                {rangeLabel}
+            </styles.sliderLabel>
+            <styles.sliderP>
+                {currentValue} of {rangeMax}
+            </styles.sliderP>
+            <styles.sliderInput
```
AbhinavMV, line 29: Range label and this displayed value can be optional....

packages/core/src/lib/Slider/types.ts
```diff
@@ -0,0 +1,7 @@
+export interface ISliderProps {
+    rangeId: string;
+    rangeLabel: string;
+    rangeMax: number;
+    rangeMin: number;
+    rangeValue: number;
+}
```
AbhinavMV, line 7: Please add comments for each type it is used by storybook docs..

packages/core/src/lib/Slider/Slider.styles.ts
3scava1i3r: will do

packages/core/src/lib/Slider/Slider.tsx
3scava1i3r: changed name also directly using them now

packages/core/src/lib/Slider/Slider.tsx
3scava1i3r: Would you mind explaining more on this,like what other than range label

AbhinavMV: Label and the displayed value i.e `{currentValue} of {rangeMax}`
User may not want to show it this way....
Instead you can make this optional using a new type probably.. if user wishes to see the message they can.
This can be added in future iterations so not a big blocker.

packages/core/src/lib/Slider/Slider.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+const Slider: React.FC<ISliderProps> = ({
+    rangeId,
+    rangeLabel,
+    rangeMax,
+    rangeMin,
+    rangeValue,
+}) => {
+    const [currentValue, setCurrentValue] = useState(rangeValue);
+
```
AbhinavMV, line 15: Let's not have an internal state instead let user pass a state variable. rangeValue could be used for that.
And then it also needs an onChange callback where user will provide the `setState` 
Something like below would be helpful.. 
```
const Comp = () =>{
  const [range,setRange] = useState(0);
  return <Slider value={range} onChange={(e)=>setRange(e.target.value)/>}
}
```

3scava1i3r: will do

3scava1i3r: @AbhinavMV the currentValue is for the input slider to understand the currentvalue, although i have added onChange callback for working on the same and getting back the value they need

packages/core/src/lib/Slider/Slider.tsx
3scava1i3r: Removing label at the moment,label could be made by using onChange callback and reieveing the value

3scava1i3r: Would implement a label props later to takle this

packages/core/src/lib/Slider/Slider.test.tsx
3scava1i3r: will do

packages/core/src/lib/Slider/Slider.tsx
AbhinavMV: @3scava1i3r so I understand what you are saying but this is adding additional complexity which is not needed...(We have this problem in many components)

To understand the problem with this do the following things:
1. Go to storybook Forms/Slider/Default.
2. Now change the rangeValue prop from the storybook... you will notice that even after changing the value in storybook the component has not been updated.

Now the solution to this is simple. In the component, we can add-
```
useEffect(()=>{setCurrentValue(rangeValue ?? 0)},[rangeValue])
```

But you see the complexity increasing right... just because we are maintaining an internal state.
So a better approach would be to completely remove the internal state and let users control it themselves.
Hope I was able to explain properly. Feel free to ping me if still anything is unclear 😄

packages/core/src/lib/Slider/Slider.test.tsx
```diff
@@ -0,0 +1,181 @@
...
+        expect(input).not.toBeNull();
+    });
+
+    it('renders the component with step value passed', () => {
+        const input: HTMLInputElement | null = container.querySelector(
+            `[data-testid="${testInputId}"]`,
+        );
+        expect(input?.step).not.toBeNull();
+        expect(input?.step).toBe(testStepValue)
+    });
```
AbhinavMV, line 135: This test is failing.
You can run `pnpm vitest --ui`  to run all tests and it will give you a UI which shows all tests and it's status

packages/core/src/lib/Slider/Slider.tsx
```diff
@@ -0,0 +1,50 @@
...
+    const rangeChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+        const newRangeValue = Number(event.currentTarget.value);
+        setCurrentValue(newRangeValue);
+        onChanges && onChanges(event);
+        console.log(currentValue);
+    };
+
+    return (
+        <DivStyled data-testid="test-input">
+            <LabelStyled data-testid="test-input-label" htmlFor={rangeId}>{rangeLabel}</LabelStyled>
```
AbhinavMV, line 29: For `data-testid` our naming convention is `test-<component name>` 
Please update all the tags likewise 😄

packages/core/src/lib/Slider/types.ts
```diff
@@ -0,0 +1,40 @@
...
+    rangeMin: number;
+    /**
+     * Pass for setting current value 
+     */
+    rangeValue?: number;
+    /**
+     * Pass for setting Changed event 
+     */
+    onChanges?: (event: React.ChangeEvent<HTMLInputElement>) => void;
+    /**
```
AbhinavMV, line 28: let use `onChange` as we use the same in all the components 

In comments you can write `callback to onchange event`

packages/core/src/lib/Slider/types.ts
```diff
@@ -0,0 +1,40 @@
...
+    currentLabel?: string | JSX.Element;
+    /**
+     * A boolean for working or not
+     */
+    disabled?: boolean;
+    /**
+     * Set the discrete step size of the element
+     */
+    Step?: number;
+}
```
AbhinavMV, line 40: please start props in lowercase 😅

AbhinavMV (changes requested):

AbhinavMV:
Hey @3scava1i3r 
The slider component works but sorry I have added some comments to make the component code similar to rest of the components 😅 
Feel free to ping me on discord or slack if you need me to explain anything again 🙌🏻 

Also I have deleted the changeset file and created a new one.. since we are doing patch release instead of minor.

AbhinavMV:
Closing this and creating a new PR since the git history is completely messed up now 😅

Issue #836, opened 2022-10-16
Issue in DatePicker Static Date when changed type from date to month


nikita12c:
In Datepicker component inside Forms section, 
When type is initially set to 'date' and static value is set then it works fine .  Ref below img :
![initialDateStatic](https://user-images.githubusercontent.com/66167434/196046949-413290b4-343f-433a-a860-27fa8ebbe012.png)

But when the type is set to month, it doesn't works instead shows blank line. This is because default value is set as 'yyyy-mm-dd' which is not acceptable for type=month. For month type it must be 'yyyy-mm'.
![afterMonth](https://user-images.githubusercontent.com/66167434/196047089-c72e65f3-e88f-479a-a2cd-6fb0b3d715a7.png)
 
So I think a new story must be created here for the case of type-month with 'yyyy-mm' as a static value to let the users know that both static date format is different for type month.

Should I create a PR for this ?

AbhinavMV:
TY 🥳 
Closed as completed by #837

PR #837, opened 2022-10-16
added story for static month in DatePicker

nikita12c:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description
In Datepicker component in Static Date story inside Forms section, when the type is set to month, it doesn't works instead shows blank line. 

Related issue: #836

### Solution Description
when type is set to date then format for default/static value is 'yyyy-mm-dd' 
when the type is set to month static value format should be 'yyyy-mm'.
I created a new story for the case of type-month with 'yyyy-mm' as a static value to let the users know that both static date format is different for type month.
finalStory


AbhinavMV:
Great job, TY for your contribution 🥳

Issue #862, opened 2022-10-26
[Upload] support for other file types


0xTijan:
Right now the Upload component only supports images. If user uploads pdf or any other type the component breaks:

![image](https://user-images.githubusercontent.com/88556617/198068137-d09a94ab-21c4-497e-af3b-6095660916ae.png)
(I uploaded pdf)

I was thinking maybe we could fix this and add another prop for devs to be able to specify witch file types are accepted.

BillyG83:
it would be cool to show what is supported for upload, the dev would maybe want to specify this as a prop object

```
supportUploads = {
  image: true,
  pdf: false,
   ... any else?
```

also make sure to not allow anything to be uploaded this could cause a vulnerability

0xTijan:
One option is to create a object like you suggested of [discrete types](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types#types) and than create the needed string for the input attribute ``${keyValue}/*`` :
```
	supportedUploads: {
		application: boolean,
		audio: boolean,
		example: boolean,
		font: boolean,
		image: boolean,
		model: boolean,
		text: boolean,
		video: boolean
	}
```
but this would remove the freedom from devs to be able to specify exactly which files to accept (for example only .png and not .jpeg)

Since there are many of these types I propose that devs specify exactly what they want (which makes it the safest and the most customizable way I think) - [see here](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/accept#limiting_accepted_file_types). Thats why I implemented it in that way in my PR. 🙌

PR #866, opened 2022-10-27
Updated Upload Component

0xTijan:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description

Related issue: [#862](https://github.com/web3ui/web3uikit/issues/862).

### Solution Description

Updated Upload component:
- users can select only certain file types to be accepted with `acceptedFiles` prop
- can set custom text with `descriptionText` prop
- using `style` prop they can set custom CSS
- if non-image file is uploaded the component won't break
- the name of the uploaded file is written below the image/icon

image
Example of non-image file uploaded.

image
Example of custom text and custom CSS.

### DIFF
--- a/packages/core/src/lib/Upload/types.ts
+++ b/packages/core/src/lib/Upload/types.ts
@@ -1,3 +1,5 @@
+import { CSSProperties } from "react";
+
 export const colorState = ['greenLight', 'redLight'] as const;
 export type TColorState = typeof colorState[number];
 export interface IUploadProps {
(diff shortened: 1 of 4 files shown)


packages/core/src/lib/Upload/types.ts
```diff
@@ -9,8 +11,22 @@ export interface IUploadProps {
     /**
      * The callback function to be called onChange
      */
-
     onChange?: (file: Blob | undefined | null) => void;
+
+    /**
+     * Accepted file types (default: all)
+     */
+    acceptedFiles?: string;
```
AbhinavMV, line 19: Could you add this link here in the comments so other devs can know what type of values can be used in this prop - https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/accept#:~:text=The%20accept%20attribute%20takes%20as,removed%20in%20favor%20of%20file.

AbhinavMV:
Nice work @0xTijan 🥳

Issue #867, opened 2022-10-28
[LinkTo]: Add customization


AbhinavMV:
Add more customization to LinkTo component
- remove underline
- change text color
- test to use typography component to display text and ability to pass props of typography to the component

PR #869, opened 2022-11-01
[LinkTo]: Add customization

0xTijan:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description

Related issue: [#867](https://github.com/web3ui/web3uikit/issues/867).

### Solution Description

Added:
- prop to remove underline
- change color of text, underline and icon
- and used Typography.

### DIFF
--- a/packages/core/src/lib/LinkTo/LinkTo.styles.ts
+++ b/packages/core/src/lib/LinkTo/LinkTo.styles.ts
@@ -7,8 +7,9 @@ import type { LinkToProps } from './types';
 
 // styles
 type TStyleProps = Pick<LinkToProps, 'iconLayout'>;
+type LinkStyledProps = Pick<LinkToProps, 'underline'>;
 
-const LinkStyled = styled.a`
+const LinkStyled = styled.a<LinkStyledProps>`
     ${resetCSS}
     ${fonts.text}
     align-items: center;
@@ -18,13 +19,14 @@ const LinkStyled = styled.a`
     height: fit-content;
     max-width: 100%;
     width: fit-content;
+    text-decoration: ${({ underline }) => underline ? 'underline': 'none'};
 
     &:hover {
         filter: brightness(0.7);
     }
 `;
 
-const InternalLinkStyled = styled(Link)`
+const InternalLinkStyled = styled(Link)<LinkStyledProps>`
     ${resetCSS}
     ${fonts.text}
     align-items: center;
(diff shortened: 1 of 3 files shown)


packages/core/src/lib/LinkTo/LinkTo.tsx
```diff
@@ -11,6 +12,8 @@ const LinkTo: React.FC<LinkToProps> = ({
     iconLayout = 'leading',
     text,
     type = 'external',
+    underline = true,
+    textColor="",
```
BillyG83, line 16: for text color you do not need to set a default of blank string
`textColor` is better

BillyG83: `underline = true,` is good however as the previous behaviour was true so it stops any unexpected behaviour

packages/core/src/lib/LinkTo/LinkTo.styles.ts
```diff
@@ -7,32 +7,39 @@ import type { LinkToProps } from './types';
 
 // styles
 type TStyleProps = Pick<LinkToProps, 'iconLayout'>;
+interface LinkStyledProps {
+    underline: boolean;
```
AbhinavMV, line 11: could you please rename this to `isUnderline` ?
Boolean props should have this type of naming convention so it is easier to understand while using the component 😄

0xTijan: Thanks for the feedback 🙌
...renamed it 😀

packages/core/src/lib/LinkTo/types.ts
```diff
@@ -1,10 +1,12 @@
+import { TypographyProps } from "../Typography";
+
 export const layoutState = ['leading', 'trailing', 'none'] as const;
 export type TLayoutState = typeof layoutState[number];
 
 export const typeState = ['email', 'external', 'internal'] as const;
 export type TTypeState = typeof typeState[number];
 
-export interface LinkToProps {
+export interface LinkToProps extends TypographyProps {
```
AbhinavMV, line 9: @0xTijan sorry but found one more bug

All props of Typography cannot be used here you have to Pick the types that are being used..
For eg: children, copyable,iconSize,onCopy  -> these are either not being used or not needed to be added to linkTo imo

AbhinavMV: Something like this would be better 🙌🏻 
```
type TTextProps = Pick<
    TypographyProps,
    'color' | 'variant' | 'italic' | 'monospace' | 'weight'
>;
export interface LinkToProps extends TTextProps {
```

AbhinavMV: Maybe It would be a nice idea to pass children too 🤔 
for eg -> cases where an image or some custom component needed to be used as a link, what do you think? (Not needed in this MR just a thought 🙌🏻 )

PR #884, opened 2022-11-09
747 input refactor

BillyG83:
FINALLY its here

some breaking changes so lets keep old input for a little while
i need to add tests and make the stories more clear tomorrow


AbhinavMV (approved):
Looks awesome 🤩 🚀

PR #900, opened 2022-11-16
fix(BannerStrip): position relative should not use z-index, top or left

BillyG83:
Screenshot 2022-11-16 at 15 27 43

1. position fixed should not add z-index, top or left... they are not needed
2. sorted CSS and props 🔤 
3. added style prop for margin in admin


AbhinavMV (approved):
LGTM, Nice fix 🚀 

Just added changeset file for release 🙌🏻

PR #923, opened 2022-11-25
[Stepper]: Vertical orientation styles

AbhinavMV:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description

### Solution Description

image


Maki44:
The code looks good to me, But a little suggestion on the style it would be great if it looked like the stepper you created for the onboarding, instead of the right side and left side having the same width,  the left part could have smaller width same as the admin onboarding and the buttons stay only on the right side also same like on the onboarding. What do you think?

AbhinavMV:
> The code looks good to me, But a little suggestion on the style it would be great if it looked like the stepper you created for the onboarding, instead of the right side and left side having the same width, the left part could have smaller width same as the admin onboarding and the buttons stay only on the right side also same like on the onboarding. What do you think?

Nice suggestion @Maki44 
I have added custom width for the stepper, now width can be controlled to some extent 😄 
Thanks for the idea 🤝 

image

Maki44 (approved):

PR #928, opened 2022-12-01
Reafactor table 

Maki44:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description

### Solution Description

Most of the code is from the original Table just change the HTML structure I used table element.


packages/core/src/lib/TableNew/Helper.tsx
```diff
@@ -0,0 +1,80 @@
+export function paginate(
```
AbhinavMV, line 1: Please use the <Pagination/> component added to the kit.
It becomes difficult to maintain two separate components better to use the existing one, and add features to it if anything is needed 😄

packages/core/src/lib/TableNew/TableNew.stories.tsx
```diff
@@ -0,0 +1,224 @@
...
+    customTableMockData,
+} from './MockData';
+import { Loading } from '../Loading';
+import { color } from '@web3uikit/styles';
+import styled from 'styled-components';
+import { Skeleton } from '../Skeleton';
+
+import TableNew from './TableNew';
+export default {
+    title: 'Layout/TableNew',
```
AbhinavMV, line 20: This creates a new section that is not needed.
Please change this to `3.Layout/TableNew`

packages/core/src/lib/TableNew/TableNew.stories.tsx
```diff
@@ -0,0 +1,224 @@
...
+        [<TextStyled>0xe922879D89E3a8D2f2D51EC8590AEF13216f9E7a</TextStyled>],
+        [<TextStyled>0xe922879D89E3a8D2f2D51EC8590AEF13216f9E7a</TextStyled>],
+        [<TextStyled>0xe922879D89E3a8D2f2D51EC8590AEF13216f9E7a</TextStyled>],
+        [<TextStyled>0xe922879D89E3a8D2f2D51EC8590AEF13216f9E7a</TextStyled>],
+    ],
+    pageSize: 10,
+    noPagination: true,
+    cellPadding: '0px',
+    customTableBorder: '1px solid black',
+};
```
AbhinavMV, line 224: I would suggest adding a new story that shows how this new version solves the problems with the old one 😄 

It would also you understand if any more tweaks are needed.

packages/core/src/lib/TableNew/atoms/Loader.tsx
```diff
@@ -0,0 +1,33 @@
...
+const Loader: React.FC<TLoaderProps> = ({ customLoadingContent }) => {
+    return (
+        <DivSpinnerLoaderParent>
+            {customLoadingContent ? (
+                customLoadingContent
+            ) : (
+                <>
+                    <Loading
+                        spinnerType={'wave'}
+                        spinnerColor={'#B0B5BF'}
```
AbhinavMV, line 20: Please use color from styles package

packages/core/src/lib/TableNew/atoms/Loader.tsx
```diff
@@ -0,0 +1,33 @@
...
+                        size={6}
+                    />
+
+                    <Typography weight="400" variant="h3" color={'#B0B5BF'}>
```
AbhinavMV, line 24: Color from styles package please

packages/core/src/lib/TableNew/atoms/Pagination.tsx
```diff
@@ -0,0 +1,83 @@
...
+    | 'onPageNumberChanged'
+    | 'pageSize'
+>;
+interface IPaginationProps extends TTableProps {
+    pageNum: number;
+    tableData: (React.ReactNode | string)[][];
+    handleSetPageNumber: (pageNum: number) => void;
+}
+
+const Pagination: React.FC<IPaginationProps> = ({
```
AbhinavMV, line 22: Please use the existing pagination component.
folder path -> `core/src/lib/pagination`

packages/core/src/lib/TableNew/atoms/styles.ts
```diff
@@ -0,0 +1,137 @@
...
+    hoverBackgroundColor: props.hoverBackgroundColor,
+}))`
+    ${fonts.text}
+    box-shadow: 0 4px 10px rgba(48, 71, 105, 0.1);
+    background-color: ${(props) => props.tableBackgroundColor ?? color.white};
+    border-collapse: collapse;
+    width: 100%;
+    td,
+    th {
+        border-bottom: 1px #cee4f3 solid;
```
AbhinavMV, line 26: color from styles package please

packages/core/src/lib/TableNew/TableNew.tsx
```diff
@@ -0,0 +1,145 @@
...
+                header={header}
+                noPagination={noPagination}
+                pageNum={pageNum}
+                pageSize={pageSize}
+                alignCellItems={alignCellItems}
+                justifyCellItems={justifyCellItems}
+                cellPadding={cellPadding}
+                headerTextColor={headerTextColor}
+                headerBgColor={headerBgColor}
+                isScrollableOnOverflow={isScrollableOnOverflow}
```
AbhinavMV, line 120: functionality provided by `isScrollableOnOverflow` prop is missing.

image

packages/core/src/lib/TableNew/atoms/TableBase.tsx
```diff
@@ -0,0 +1,161 @@
...
+                                        ))}
+                                </DivTableCell>
+                            </th>
+                        ))}
+                    </tr>
+                </thead>
+                {!isLoading && (
+                    <tbody>
+                        {computeCurrentData().map((row, rowKey) => (
+                            <tr className={`${hover && 'hover'}`}>
```
AbhinavMV, line 119: please add a proper className to this component eg: `web3uikit-table-row`
This would give us the ability to customize outside of the kit too (might be needed for some complex use-case in admin)

Also we can remove `hover` prop.
Only `hoverBackgroundColor` should be ok. If hoverBackgroundColor has value we show that color else the default one.
What do you think ?

packages/core/src/lib/TableNew/index.tsx
```diff
@@ -0,0 +1,4 @@
+export { default as NewComp } from './TableNew';
+export type { TableNewProps } from './types';
+// above is boilerplate stuff
+// replace with your component & props
```
AbhinavMV, line 4: sorry about this but could you remove these two comments 🙈

AbhinavMV (commented):
Hey @Maki44,  nice work 
Sorry but got some comments 🙈

packages/core/src/lib/TableNew/Helper.tsx
Maki44: I didn't know, that there is a pagination component 🙈

BillyG83:
imma close this PR because we have a new responsive table in the works, but thanks for the effort @Maki44 you will always be part of the web3uiKit community 💚

PR #937, opened 2022-12-07
fix(Illustration, Modal, Selects, TextArea): remove dateNow error

BillyG83:
removing date now ID, tests all pass, i can not remember why i thought this was a good idea


AbhinavMV (approved):
LGTM 🚀 
Let's go

PR #944, opened 2022-12-12
fix(Select): traditional placeholder prop used

BillyG83:
Select with traditional mode was not using the placeholder


AbhinavMV (approved):
LGTM 🚀

Issue #951, opened 2022-12-14
[dropdown] align center and add justify content prop


dani69654:
The content should be aligned.
![Screenshot 2022-12-14 at 15 26 35](https://user-images.githubusercontent.com/60792849/207623730-6abaadb9-529e-486e-aee9-059de34305a5.jpg)

Add justify content prop so that the dropdown can look like this

![Screenshot 2022-12-14 at 15 34 31](https://user-images.githubusercontent.com/60792849/207623674-38927bae-d55c-4671-b000-696d30715ee9.jpg)

PR #954, opened 2022-12-14
enable center align for dpdw content + fix alignment

dani69654:
closes https://github.com/web3ui/web3uikit/issues/951

https://www.loom.com/share/8cbcb798db7f4156b998db424383f99e


AbhinavMV (commented):
Hey bro nice work 
Just have some questions
1. When the option with icon (eg: <icon> testnet) is selected, this is how it is displayed. Is this how it should be displayed? 🤔  
image

2. From the design in the issue I can see that even the dropdown arrow is centered. but not in the story. Maybe design was not updated not sure. 

If these two things are ok and no changes are needed to PR please go ahead and merge this @dani69654 🙌🏻

dani69654:
Hey @AbhinavMV , thanks for reviewing:

> When the option with icon (eg: testnet) is selected, this is how it is displayed. Is this how it should be displayed?
This is how it works now, not sure if it is correct, I did not work on that in this MR. If we don't like it we can surely change its behaviour.

> From the design in the issue I can see that even the dropdown arrow is centered. but not in the story. Maybe design was not updated not sure.

You are right, I missed that. On it, good catch

dani69654:
Screenshot 2022-12-15 at 10 01 39

AbhinavMV (approved):
LGTM 🚀

PR #957, opened 2022-12-16
[Wave Loading]: size prop fix 

AbhinavMV:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description

### Solution Description

Wave loader had size hardcoded. Now it is customizable using size prop.

image

### DIFF
--- a/.changeset/fluffy-ducks-own.md
+++ b/.changeset/fluffy-ducks-own.md
@@ -1,0 +1,5 @@
+---
+'@web3uikit/core': patch
+---
+
+Custom size for wave loader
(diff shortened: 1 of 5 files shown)


BillyG83 (approved):
LGTM but for animation it is much smoother when you don't mix it with JS so much

AbhinavMV:
> LGTM but for animation it is much smoother when you don't mix it with JS so much

Hmm but i don't see any other solution apart from creating the same component again in admin 😅 
Will merge this and if there is some noticeable animation lag then will remove this change with another MR

PR #961, opened 2022-12-18
Fix avatar 3 letter

BillyG83:
you can set the amount of characters to show in the Avatar


BillyG83 (commented):
perfect PRs do exist after all

BillyG83:
@rayyan224 is a believer 💚 

Screenshot 2022-12-18 at 16 11 54

PR #963, opened 2022-12-18
feat(Skeleton): custom bg colors / updated animation

BillyG83:
- added bg color prop
- added animation color prop
- updated the animation to make it more clear and customisable


BillyG83 (commented):
yes i like!

PR #969, opened 2022-12-27
[NFTCard] : New styles for card

AbhinavMV:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description

### Solution Description

image


packages/core/src/lib/NftCard/types.ts
```diff
@@ -38,4 +38,14 @@ export interface INftCardProps {
      * custom details component
      */
     customDetails: ReactNode;
+
+    /**
+     * set card background color
+     */
+    backgroundColor: string;
```
BillyG83, line 45: should this be optional?

packages/core/src/lib/NftCard/types.ts
```diff
@@ -38,4 +38,14 @@ export interface INftCardProps {
...
+
+    /**
+     * set border for details section
+     */
+    detailsBorder: string | 'none';
```
BillyG83, line 50: this seems strange. maybe it should be optional? but the dev cant even use it to set a border... If its an on / off switch use a boolean like hasBorderStyles. But best DX is to offer a `border?: string` so the dev can set `border: 1px solid color` and if its left unset it shows what you have now... or you set current styles as the default value on line 17 of `packages/core/src/lib/NftCard/NftCard.tsx`

packages/core/src/lib/NftCard/types.ts
AbhinavMV: yeah, nice catch 🙌🏻 

Will make these new props optional ✅

BillyG83 (approved):
you can merge this if you need a quick fix but consider to open an iteration task. Actually this feedback plus using the `TCustomize` type for this would be Super Saiyan level dev work

packages/core/src/lib/NftCard/types.ts
AbhinavMV: - Yes this is optional.
- It's not named as `border` because this border is for the component inside card.
- will remove extra `none` ✅

AbhinavMV:
Have added `TCustomize` instead of just background color 🙌🏻

PR #972, opened 2022-12-29
[Hero]: text color and customize prop added

AbhinavMV:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description

### Solution Description
Fixed same text color for all component issue.
Added customize prop as well.

image

### DIFF
--- a/packages/core/src/lib/Hero/Hero.tsx
+++ b/packages/core/src/lib/Hero/Hero.tsx
@@ -12,7 +12,7 @@ const {
 
 const Hero: React.FC<IHeroProps> = ({
     align = 'center',
-    backgroundColor = `${color.gray30}`,
+    backgroundColor,
     backgroundURL = '',
     children,
     customImage = {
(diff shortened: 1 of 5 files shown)


packages/core/src/lib/Hero/Hero.tsx
```diff
@@ -3,25 +3,30 @@ import { color } from '@web3uikit/styles';
...
+    ImageStyled,
+    LeftContainerDiv,
+    RightContainerDiv,
+    SectionStyled,
+} = styles;
 
 const Hero: React.FC<IHeroProps> = ({
     align = 'center',
-    backgroundColor = `${color.gray30}`,
+    backgroundColor,
```
AbhinavMV, line 15: these styles can also be set via customize prop, and if these default values are kept those styles are not set.

Instead, these default values are set directly in the styles file

for eg:
```
 background-color: ${(p) => p.backgroundColor ?? p.customize?.backgroundColor ?? color.gray30};
```

BillyG83 (approved):
niiiiiice TCustomize will help align props things as we go forward 👍

PR #983, opened 2022-12-31
[Modal,NftCard,Select,Input,]: Bug fix and feature added

AbhinavMV:
---
name: 'Pull request'
about: A new pull request
---

### Checklist

### Issue Description and Solution Description

1. Modal 
- added customize prop

Solution:
image

2.  NFTCard 
- for palm network the images are sometimes in `metadata.file` in api result, added fallback image if there is any issue in loading ipfs image (images don't load on palm network sometimes)

Solution: Added more conditions in image display logic

3. Select (Beta) 
- When select menu is opened the first element has active hover style, now this is removed.
- Reset active element hover style and menu scroll every time select is closed/opened.
- add `fixed` to overlay element

Issue:
image

4. InputNew 
- Add `tabIndex={-1}` to password button to avoid it on tab click.
- sync error message for react hook form

Issue: when used with react hook form, the error message is updated after first render but its not updated in the input component
image

5. Add new color - purple-60


packages/core/src/lib/Select/SelectBeta/SelectBeta.styles.ts
```diff
@@ -167,7 +167,7 @@ const DivStyledOverlay = styled.div`
     bottom: 0;
     display: none;
     left: 0;
-    position: absolute;
+    position: fixed;
```
BillyG83, line 170: is this correct? wont the overlay be fixed to the page instead of the parent? does this casue an issue when you scroll the page?

packages/core/src/lib/InputNew/atoms/VisibilityToggle.tsx
```diff
@@ -12,6 +12,7 @@ const VisibilityToggle: React.FC<IVisibilityToggleProps> = ({
         className="input_visibility"
         data-testid="test-input-icon-visibility"
         onClick={() => onClick()}
+        tabIndex={-1}
```
BillyG83, line 15: why minus -1?

packages/core/src/lib/Select/SelectBeta/components/SelectMenuList.tsx
```diff
@@ -68,7 +68,7 @@ const SelectMenuList: React.FunctionComponent<ISelectExtendedProps> = ({
         if (wrapper) {
             const el = wrapper.querySelector(selector);
             if (el) {
-                el.scrollIntoView(false);
+                el?.scrollIntoView?.(false);
```
BillyG83, line 71: i didnt know `scrollIntoView` was a function now, we used to do this manually. It seems well supported but there is some notes about Safari

https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView

BillyG83 (approved):
some small questions more than comments. Also it would be nicer to break something like this into smaller MRs if possible 👍

packages/core/src/lib/Select/SelectBeta/components/SelectMenuList.tsx
AbhinavMV: the function is supported only the options parameter is not, just to be on the safer side I added optional chaining too.
I tested it on safari and it is working 👍🏻

packages/core/src/lib/InputNew/atoms/VisibilityToggle.tsx
AbhinavMV: on password type input, when you click tab it focuses on the password icon to avoid this negative index is being used. (Never seen any site implement focusing on the password icon)

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex
```
A negative value (usually tabindex="-1") means that the element is not reachable via sequential keyboard navigation, but could be focused with JavaScript or visually by clicking with the mouse. It's mostly useful to create accessible widgets with JavaScript.
```

packages/core/src/lib/Select/SelectBeta/SelectBeta.styles.ts
AbhinavMV: the overlay is used to close the select menu if someone clicks outside of the select.

If absolute is used it takes the parent size which is usually not the screen size so the menu won't be closed. (checkout select component in admin playground it won't close even if clicked outside)

But if fixed is used it take up entire screen and if user clicks anywhere the select component will be closed

AbhinavMV:
yeah sorry, my bad I thought I was working on different branches but only realized that was not the case after everything was done.

Will take care in future 🙌🏻

BillyG83 (approved):
🚢 Ship it!

PR #1015, opened 2023-01-23
1008 select html5 default index

BillyG83:
Select html5 can take default index of 0, stupid Javascript!


BillyG83:
@rayyan224 approved this 😅

PR #1038, opened 2023-02-26
Select HTML5 & InputNew fixes

BillyG83:
- Select HTML5 controlled by state meaning if the value prop changes the value inside it is updated
- Input New controlled by state meaning if the value prop changes the value inside it is updated
- minor style fixes for input and select in confirm and error state when working with input_filled className

### DIFF
--- a/packages/core/src/lib/Select/TraditionalSelect/TraditionalSelect.tsx
+++ b/packages/core/src/lib/Select/TraditionalSelect/TraditionalSelect.tsx
@@ -11,7 +11,7 @@ const TraditionalSelect: React.FC<ISelectProps> = ({
     description,
     disabled = false,
     errorMessage = '',
-    id,
+    id = 'web3uikit-select',
     label,
     onBlurTraditional,
     onChange,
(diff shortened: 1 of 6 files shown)


BillyG83 (commented):
annoying it looks like we need to use start to programatically update values

packages/core/src/lib/Select/TraditionalSelect/TraditionalSelect.tsx
```diff
@@ -30,48 +30,65 @@ const TraditionalSelect: React.FC<ISelectProps> = ({
...
                 required={validation?.required}
-                value={value}
+                value={currentValue || placeholder || 'Please choose'}
             >
                 <option disabled>{placeholder || 'Please choose'}</option>
 
                 {options.map((option, i) => (
                     <option
                         data-testid={`test-select-option-${i}`}
+                        disabled={option.disabled || false}
```
AbhinavMV, line 81: there is no style indicating that the component is disabled. 

image
This is the disabled component and looks like it is not disabled. but on clicking it nothing happens. I think we do have designs for it

AbhinavMV:
If the above comment is not a big issue for your requirement then we can go ahead and merge this PR and add the style later.
LGTM 🚀

AbhinavMV (approved):

BillyG83:
> If the above comment is not a big issue for your requirement then we can go ahead and merge this PR and add the style later. LGTM 🚀

its a very good catch, yeah i also noticed that the outline didnt change. I was trying to debug it but did you notice that Storybook flips out when you open the console? I'll take another look soon. I think the wrapper (outline) of all form elements are the same. I'd love to get some time to break them down and make them more reusable

PR #1045, opened 2023-03-14
1043 tooltip customisable

BillyG83:
- Tooltip take custom props
- Tooltip takes custom JSX
- Tooltip stays on hover so user can interact with JSX

Screenshot 2023-03-14 at 14 54 10

Screenshot 2023-03-14 at 15 09 10

### DIFF
--- a/.changeset/polite-tools-happen.md
+++ b/.changeset/polite-tools-happen.md
@@ -1,0 +1,5 @@
+---
+'@web3uikit/core': patch
+---
+
+fix(tolltip): added story to show custom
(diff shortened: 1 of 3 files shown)


AbhinavMV (approved):
LGTM 🚀

BillyG83:
> LGTM 🚀

Sorry i didnt tag you man, small fixes like this is below your pay grade now 😆 

but thanks for the feedback. You will always be web3uiKit King even when you are 10x BE dev

PR #1046, opened 2023-03-14
feat(Notification): extended to use TCustomize prop

BillyG83:
- using TCustomize prop
- horrible story to showcase all feature
- takes JSX as content
- all clickable with with onHover feature we prop wont use
- 
Screenshot 2023-03-14 at 16 22 50


KristamMoffett (approved):
LGTM! No comments :)

PR #1051, opened 2023-03-22
Fix(logo): Add color to visa amex logos

johanMoralis:
---
name: 'Fix(logo): Add color to visa amex logos
about: Visa and Amex logo no longer only white
---

### Checklist

### Issue Description

Visa and Amex logos are white and don't support other colors. This adds the possibility to add color to those logos. 


AbhinavMV (approved):
@johanMoralis LGTM 🚀 

QQ: Is the same not required for diners, mastercard etc ?

johanMoralis:
> @johanMoralis LGTM 🚀
> 
> QQ: Is the same not required for diners, mastercard etc ?

Diners and Mastercard already have colors in their logos by default which means there is no reason to change between for example white and blue depending on background. Mastercard is always yellow/orange with light or dark background.

PR #1053, opened 2023-03-28
fix(Select): html respects value passed

BillyG83:
Default (no value)
Screenshot 2023-03-28 at 12 15 40

Value passed as string into value prop
Screenshot 2023-03-28 at 12 15 48

Value being passed as default index
Screenshot 2023-03-28 at 12 15 54


KristamMoffett (approved):
Yep! LGTM!

PR #1054, opened 2023-03-28
feat(Toggle): added component

BillyG83:
added component

it has noral HTML events plus `onToggle` to get the value back as string for on and off

off / left
Screenshot 2023-03-28 at 22 47 13

on / right
Screenshot 2023-03-28 at 22 47 18

customize
Screenshot 2023-03-28 at 22 47 27
Screenshot 2023-03-28 at 22 47 36


AbhinavMV (approved):
LGTM 🚀 

Noticed 1 thing that design for this are a bit different.  the toggle button is a bit different in design. I guess its not a blocker right now ?

image

PR #1063, opened 2023-04-07
fix(plancard): Added back footer and topLabel does no longer contain …

johanMoralis:
---
Update to Placard
Added back footer and topLabel does no longer contain
---

### Checklist

### Issue Description

Footer was removed but was still defined as a type in the component. I have now added back footer so it works again.  
TopLabel previously contained a condition to show checkmark if it was the users current billing period of that plan. I have removed this condition and only kept the top label now so there is more flexibility when implementing the placard.

### DIFF
--- a/.changeset/fast-hornets-peel.md
+++ b/.changeset/fast-hornets-peel.md
@@ -1,0 +1,5 @@
+---
+'@web3uikit/core': patch
+---
+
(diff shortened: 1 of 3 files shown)


AbhinavMV (approved):
LGTM 🚀

